PR #4, opened 2015-09-10
Shuffle/unshuffle loops

LucianMarc:
Settled groups of instructions in loops. Passes the bench suite in the cross-compiler.


FrancescAlted:
Very nice vectorization :+1:

PR #6, opened 2016-02-17
Initial support for zstd

FrancescAlted:
Not ready for merging yet.  Once ztsd contexts are used, we will be able to be merged.  Dictionaries will probably come at a later stage.


FrancescAlted:
As dealing with zstd contexts requires a context refactorization in blosc itself, let's merge this first.

PR #14, opened 2017-03-26
Make blosc2 build without zstd

jcftang:
This just adds in some missing ifdef's I needed this to build this https://github.com/jcftang/go-blosc/commit/8be85d0a64164f1586021beba2177cae98b38916


FrancescAlted:
Sure thing.  Thanks!

PR #16, opened 2017-05-26
Fixed docstrings in blosc.h

albertosm27:


FrancescAlted:
Looks good.  Maybe you can further indent the comments on the right part of code lines.  Thanks!

albertosm27:
The comments were alright, there are too long for going on the side with the 80 column style.

PR #47, opened 2018-11-03
omit msvc warnings, mostly by casting to the required types

stoni:
* By applying these changes all warnings from the Visual Studio Compiler (2017) are gone
* Most of the warnings are due to type mismatches.
* Most of the mismatches are not problematic in real world applications, however it always better to avoid the warnings by resolving the type conflicts without casting.

### DIFF
--- a/blosc/frame.c
+++ b/blosc/frame.c
@@ -26,6 +26,8 @@
     #include <stdint.h>
   #endif
 
+  #define fseek _fseeki64
+
 #endif  /* _WIN32 */
 
 /* If C11 is supported, use it's built-in aligned allocation. */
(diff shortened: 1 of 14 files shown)


blosc/frame.c
```diff
@@ -671,13 +673,15 @@ void* blosc2_frame_append_chunk(blosc2_frame* frame, void* chunk) {
   int64_t frame_len;
   int64_t nbytes;
   int64_t cbytes;
-  uint32_t chunksize;
+  uint32_t ichunksize;
```
FrancescAlted: Maybe `uchunksize` should be a better name for unsigned ints?

FrancescAlted:
I like this (bar some names of variables, which I can fix later on).  Thanks!

blosc/frame.c
stoni: sure, renaming would make sense. I'm sure there is a reason, but ideally the chunksize would be the same datatype everywhere ;)

PR #60, opened 2019-06-12
Support for prefilters

FrancescAlted:
Prefilters are user-provided functions that are meant to be executed previously to any filter. As these functions are executed in the same pipeline than the filters, the memory usage is optimal; in addition, we get parallelism for free.


FrancescAlted:
All tests pass in Linux, but the `test_prefilter` still fails on Win.  As I don't have a Windows box handy, I have gave a shot at the test with valgrind, but nothing bad was revealed.  Merging and will address the Windows issue later on.

PR #75, opened 2019-08-27
Metalayers moved to schunk

FrancescAlted:
Metalayers should be able to be attached into superchunks, and not only to frames.  This is a attempt in order to move them into the schunk object.  It seems to me that the refactoring leads to a more consistent API too, so even if it is a breaking change, I think the benefits are worth it (we are in beta after all).


FrancescAlted:
This is in a much better state now, and ready for test.

PR #83, opened 2019-09-11
const-qualify static strings

stevengj:
This is a minor PR that changes `char*` declarations to `const char*` wherever they refer to static strings or string literals.  This is useful for two cases especially:

* In a function like `blosc2_frame_from_file(const char *fname)` that takes a string as an argument, it makes it explicit that Blosc does not take "ownership" of the pointer (it internally allocates its own copy if needed in all such functions that I changed).  Also, it eliminates [deprecation warnings in C++](https://stackoverflow.com/questions/13690306/conversion-from-string-literal-to-char-is-deprecated) if one passes a string literal for such arguments.

* In a function like `const char* blosc_get_version_string(void)`, it makes it clear that the caller should not call `free` on the returned pointer.

### DIFF
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -206,7 +206,7 @@ static int compname_to_clibcode(const char* compname) {
 }
 
 /* Return the library name associated with the compressor code */
-static char* clibcode_to_clibname(int clibcode) {
+static const char* clibcode_to_clibname(int clibcode) {
   if (clibcode == BLOSC_BLOSCLZ_LIB) return BLOSC_BLOSCLZ_LIBNAME;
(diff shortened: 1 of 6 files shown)


stevengj:
Note that some of these changes were apparently also made in c-blosc by Blosc/c-blosc#218, but were not yet ported over to c-blosc2?

FrancescAlted:
@stevengj Yes, some of the changes in C-Blosc have not been ported to C-Blosc2 yet.  This was not intentional, but rather just lack of time.  Thanks for taking care of this.

FrancescAlted:
Thanks!

PR #89, opened 2019-09-25
Shuffle implemented using altivec vector style instructions

kif:
This is work in progress but the first function has been translated and apparently works.
Are there any guidance for the style ? I have no experience in C-programming, maybe there are some stupid errors done here and there.

### DIFF
--- a/blosc/shuffle-altivec.c
+++ b/blosc/shuffle-altivec.c
@@ -68,16 +68,16 @@ static inline __vector uint8_t helper_misaligned_read( int32_t position,
                                                        const uint8_t * src,
                                                        __vector uint8_t * previous,
                                                        __vector uint8_t shuffle){  
-  __vector uint8_t vec0, vec1, data;
+  __vector uint8_t xmm0, xmm1, data;
   // Insert our data into the low and high vectors
   if (! previous)
-    vec0 = vec_ld(position, src);
+    xmm0 = vec_ld(position, src);
   else
-    vec0 = previous[0];
-  vec1 = vec_ld(position + 16, src);
-  data = vec_perm(vec0, vec1, shuffle);
+    xmm0 = previous[0];
+  xmm1 = vec_ld(position + 16, src);
+  data = vec_perm(xmm0, xmm1, shuffle);
   // Store only the high part in the previous buffer to re-use in next read.
-  previous[0] = vec1;
+  previous[0] = xmm1;
   // and return the data 
   return data;
 }
@@ -91,8 +91,6 @@ static inline __vector uint8_t helper_misaligned_read( int32_t position,
  *     previous: Pointer to the remaining vector of the last write: read and write !
  *     shuffle_lo: the vector for shuffling the lower part
  *     shuffle_hi: the vector for shuffling the higher part
- * 
- * return: None
  */
 static inline __vector uint8_t helper_misaligned_store(__vector uint8_t data, 
                                                        int32_t position,  
@@ -154,17 +152,24 @@ static inline __vector uint8_t gen_permute_read(uint8_t offset){
   return permute;
 } 
 
+static inline __vector uint8_t gen_save_mask(uint8_t offset){
+  __vector uint8_t mask;
+  for (uint8_t k = 0; k < 16; k++)
+    mask[k] = (k<offset)?0:0xFF;
+  return mask;
+}
+
 
 /* Routine optimized for shuffling a buffer for a type size of 2 bytes. */
 static void
 shuffle2_altivec(uint8_t* const dest, const uint8_t* const src,
                  const int32_t vectorizable_elements, const int32_t total_elements){
-  const uint8_t bytesoftype = 2;
+  static const uint8_t bytesoftype = 2;
   const uint8_t roffset = ((size_t) src) % 16;
   static __vector uint8_t perm_lo[2], perm_hi[2], perm;
   uint32_t j;
   uint8_t i;
-  __vector uint8_t vec0[2], vec1[2], low[2], lor;
+  __vector uint8_t xmm0[2], xmm1[2], low[2], lor;
   uint8_t woffset[2];
   uint8_t* desti[2];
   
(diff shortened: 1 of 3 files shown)


blosc/shuffle-altivec.c
```diff
@@ -15,61 +16,145 @@
...
-          buf[8], buf[9], buf[10], buf[11],
-          buf[12], buf[13], buf[14], buf[15]);
+    //Load the surrounding area
+    __vector uint8_t low = vec_ld( position, dst);
+    __vector uint8_t high = vec_ld( position+16, dst);
+    //Prepare the constants that we need
+    uint8_t offset = ((size_t)dst) % 16;
+    __vector uint8_t permute_lo, permute_hi;
+    for (uint8_t i=0; i<16; i++)
+    {
```
FrancescAlted: I typically prefer the curly brace to be at the end of `for` statement (kind of similar to the colon at the end of a `for` in Python).

blosc/shuffle-altivec.c
```diff
@@ -15,61 +16,145 @@
...
-
-/* The next is useful for debugging purposes */
-#if 0
+#include <altivec.h>
 #include <stdio.h>
-#include <string.h>
 
-static void printxmm(__m128i xmm0)
+
+//Store a vector to an unaligned location in memory
```
FrancescAlted: Add a space between the '//' and the actual comment (similar to the space after the ' #' comment in Python).

blosc/shuffle-altivec.c
```diff
@@ -15,61 +16,145 @@
...
+            permute_hi[i] = 32 + i - offset;
+        }
+        else{
+            permute_lo[i] = 16 + i - offset;
+            permute_hi[i] = i;
+        }
+    }
+    //Insert our data into the low and high vectors
+    low = vec_perm( low, v, permute_lo );
+    high = vec_perm( high, v, permute_hi );
```
FrancescAlted: I typically don't add spaces at the beginning or the end of parameter list in functions.

blosc/shuffle-altivec.c
```diff
@@ -15,61 +16,145 @@
...
+} 
+
+/* Calculate the permutation vector for storing data in the upper part
+ * 
+ * param: offset, the position
+ */
+static inline __vector uint8_t gen_permute_high(int32_t offset)
+{
+    __vector uint8_t permute;
+    for (int32_t i=0; i<16; i++)
```
FrancescAlted: Use spaces between variables and assignements or comparisons, just like Python does.

blosc/shuffle-altivec.c
```diff
@@ -15,61 +16,145 @@
...
+    /* Transpose vectors */
+    out0 = vec_perm(inp0, inp1, even);
+    out1 = vec_perm(inp0, inp1, odd);
     /* Store the result vectors */
-    dest_for_jth_element = dest + j;
-    for (k = 0; k < 2; k++) {
-      _mm_storeu_si128((__m128i*)(dest_for_jth_element + (k * total_elements)), xmm1[k]);
+    vec_st(out0, j, dest);
+    if (offset)
+    {// misaligned store
```
FrancescAlted: The opening curly brace should go up.  OTOH, the closing brace is recommended to be in their own line.

blosc/shuffle-altivec.c
```diff
@@ -579,6 +725,17 @@ unshuffle_altivec(const int32_t bytesoftype, const int32_t blocksize,
     return;
   }
 
+  /* Warning in case of unaligned buffers*/
+  if (((size_t)_dest) % 16)
+  {
+       printf("shuffle2_altivec: Unaligned destination !\n");
```
FrancescAlted: Be consistent and use 2 spaces for indentation (the convention for Blosc2 code).

kif: I'll setup the editor accordingly. thanks

kif:
Thanks Francesc for the comments. 
I wonder if there is any risk of having the provided buffers miss-aligned (on a 16-byte boundary) ?
Are you doing anything specific with those buffers before the sending them to the shuffling ?

FrancescAlted:
Yes, you should be careful about using primitives that allow to read from miss-aligned buffer ([here it is how it is done in SSE2 code](https://github.com/Blosc/c-blosc2/blob/master/blosc/shuffle-sse2.c#L52)). And the same goes for writing to user-provided buffers.  I did not detect a noticeable loss in performance when using `_mm_loadu_si128` or `_mm_storeu_si128` instead of their aligned counterparts.  Finally, providing tests where you pass miss-aligned buffers is useful.

kif:
As I am already managing the misaligned write for the most of the store, I can easily extend it to all store actions.
My the way, I performed an initial benchmark to compare the translated SSE2 with the hand-written Altivec (VSX to make IBM people happy):
image
image

kif:
I got the confirmation by IBM that there is no guarantee any buffer should be aligned. There is hence more work to do on the read side.

FrancescAlted:
Hi @kif .  Is this ready to be merged?  This looks pretty good to me already.

kif:
Hi Francesc,

The shuffle part is OK, close to optimal.
I am working on the bitshuffle part. It is up to you f you prefer one
PR or wish to wait for the two filters to be ready.

Cheers,

-- 
Jérôme Kieffer
tel +33 476 882 445

FrancescAlted:
No, waiting is fine.  Just tell me when you think the PR would be ready.  Thank you!

kif:
Hi Franscesc,
I have a question about bitshuffle and the way it is implemented in SSE2 (vs Neon):
I noticed some severe performance drop for bitunshuffle when the size of the bloc goes from 512k to 1M (when run on the power9 with translated intrinsic instructions, the drop appears earlier on x86). I have the feeling that this is due to the usage of a temporary buffer to store the data of the byte-shuffling before the bitwise transpose within a vector.

Did you try to load a bunch of vectors (8, 16, 32, 64, 128 depending on the element size) on the stack, perform the byte-transpose then the bit-transpose. I propose an implementation here: 
https://github.com/Blosc/c-blosc2/pull/89/files#diff-5277a43f9b29dffae0360250cb83dc65R159
but did not benchmark it.
It gives the proper results but does not yet handle the end of the buffer for the last 15 elements.
I think this is the path you followed for the neon implementation  ... but you mentioned once the neon code is not in production due to some bugs ... what were those bugs actually ?
thanks

FrancescAlted:
Yeah, so the bugs for bitshuffle for NEON are uncovered when you activate the tests in the branch [bitshuffle-tests](https://github.com/Blosc/c-blosc2/compare/bitshuffle-tests).  I am not the original author of the code (@LucianMarc was), but I'd say that it should be relatively easy to fix the remaining issues, although I have not put the time for doing that yet (help is welcome).

Regarding the drop in performance from 512k to 1M, my suggestion is indeed to actually benchmark your alternative solution for a number of block sizes and see how it goes.

kif:
I can confirm you the approach without third buffer for  the bitunshuffle performs much better (here implemented on 8-bits data)
image

FrancescAlted:
Hi @kif .  So it looks like performance is really good and tests are passing.  Also, and after a quick glance at the code, it looks good to me.  I assume that this is ready for a merge, but tell me if you want more time for fixing some details.

kif:
Hi Francesc,

Yes, the results are good, but the implementation is currently limited to 8-bit data-type and does not handle the remaining bytes (when the buffer size is not a multiple of the bloc size).

The graphic provides figures of how much is still to win on this side (the orange curve is production ready, the red one not yet, a factor 2 at best). I had limited time to work on it recently as we discovered the largest bottleneck was in the h5py implementation to access our data. As there are low-hanging fruits over-there I have been invited to focus more on h5py.

I believe there is less than a week of work to finish this. Would you consider providing us some consultancy and/or help to make this working at full speed ? (the conversation should go on privately then).

Feel free to merge as it works as it is.

Cheers,

FrancescAlted:
Ok, merged then.

For the record, I have sent you a private message about consultancy services.

PR #99, opened 2020-01-10
Enable ALTIVEC extension regardless to the version of gcc

kif:
Tested on gcc7 where the option NO_WARN_X86_INTRINSICS was not (yet) present (appeared in gcc8). Did not manage to have *clang* compiling the same way :(


FrancescAlted:
LGTM.

PR #102, opened 2020-03-23
New pparams structure convention

FrancescAlted:
This is a new convention for the `pparams` struct so that it introduces a more generic mechanism for prefilter functions to gather the necessary info from it.


FrancescAlted:
Only one platform in CI ([Windows, Visual Studio 2015, 32-bit](https://ci.appveyor.com/project/FrancescAlted/c-blosc2/builds/31651351)) is crashing on a single test (and it varies from run to run), and reason is not clear at all (and most probably unrelated with this PR) so merging.

PR #109, opened 2020-04-29
Cleanup CMake

nmoinvaz:
These are changes that clean up the cmake files so it is more readable and style is standardized.


FrancescAlted:
LGTM

Issue #96, opened 2019-12-17
Disable C++ compiler detection 


FrancescAlted:
Ditto in https://github.com/Blosc/c-blosc/issues/284

nmoinvaz:
I believe this issue can be closed now.

FrancescAlted:
Right.  Thank you.

PR #110, opened 2020-04-29
Disable C++ compiler detection.

nmoinvaz:
See #96. This should speed up CI build time a bit, because cmake doesn't have to detect C++ compiler on each machine.


FrancescAlted:
LGTM

PR #119, opened 2020-06-15
fix neon implementation of bitshuffle

cmulter:
Bitshuffling is broken in the neon implementation.
This PR fixes the missing return path. The shuffling function now always returns the shuffled size.

### DIFF
--- a/blosc/bitshuffle-neon.c
+++ b/blosc/bitshuffle-neon.c
@@ -951,6 +951,8 @@ bitshuffle_neon(void* _src, void* _dest, const size_t size,
          so we're done processing here. */
       return count;
   }
+
+  return (int64_t)size * (int64_t)elem_size;
 }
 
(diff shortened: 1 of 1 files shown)


FrancescAlted:
Cool!  With that tests are passing on ARM?  We would need to setup some CI in ARM.  I think GitHub actions is providing support for this platform.

nmoinvaz:
GHA doesn’t support ARM out of the box. You have to use qemu. In zlib-ng we cross compile and then use qemu with cmake.

cmulter:
> Cool! With that tests are passing on ARM? We would need to setup some CI in ARM. I think GitHub actions is providing support for this platform.

All test are passing on arm64 with neon.
I am not familiar with CI in github, so i cannot help with that.

FrancescAlted:
Fair enough.  Thank you!

PR #120, opened 2020-06-25
Group third-party source files with CMake

nmoinvaz:
Groups files into folders in project file. when building for Xcode it should do something similar. Here is for VStudio..
image


FrancescAlted:
LGTM

PR #126, opened 2020-06-26
Fixed missing blosc_cbuffer_validate function

nmoinvaz:
This also fixes the following ASAN error: https://oss-fuzz.com/testcase-detail/5114316075565056.


FrancescAlted:
LGTM.

PR #129, opened 2020-06-28
Prefilters memcpyed

FrancescAlted:
This allows to activate prefilters when either the compression is deactivated (`clevel=0`) or when regular compression fails and the data is just `memcpyed`.


FrancescAlted:
The platforms that are failing look like a temporary failure in dowloading dependecies.  The only platform that is missing is Win32, but this seems unrelated with this (see #133).  I say let's merge this.

PR #131, opened 2020-06-30
Delete development environment config files that might conflict

nmoinvaz:
This will help me as switching between branches and various git resets sometimes my config files are being wiped out.


FrancescAlted:
LGTM

PR #142, opened 2020-07-29
using compiler build in function

jinfeihan57:
It should fast.

### DIFF
--- a/blosc/blosc-private.h
+++ b/blosc/blosc-private.h
@@ -24,17 +24,15 @@ extern "C" {
 
 // Return true if platform is little endian; else false
 static bool is_little_endian(void) {
-  bool little;
-  int i = 1;
+  const static int i = 1;
   char* p = (char*)&i;
 
   if (p[0] == 1) {
-    little = true;
+    return true;
   }
   else {
-    little = false;
+    return false;
   }
-  return little;
 }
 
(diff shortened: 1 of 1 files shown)


FrancescAlted:
Yes, it is possible that your proposal is faster.  On the other hand, it also makes the code slightly more difficult to understand.  As a rule of thumb, we would gladly accept this sort of micro-optimizations if they prove to be useful in real cases.  IMO, the proposed code here is just (perhaps) accelerating metadata handling, so I don't think this would have a measurable effect in general (but I'd love to be proved wrong).

jinfeihan57:
The function be called many times.
so I test it. 
using bench/b2bench.exe blosclz 
Alternate test 10 times：
Elapsed time :
                         15206.3   15719.4
                         14544.6   15630.8
                         15239.6   15354.9
                         15235.0   15389.0
                         15074.4   15477.4
                         15096.4   15471.3
                         15307.5   15351.6
                         14832.0   15706.0
                         15165.3   15407.7
                         15250.6   15465.5
Average time：15095.17 15497.36
400/15100 = 2.65%
Improved by 2.65%

FrancescAlted:
Yeah, I can reproduce the general increase of ~2% in performance.  Ok, let's merge this.  Thanks for the extra benchmarking time!

PR #143, opened 2020-08-02
Minor changes to the frame document for readability.

nmoinvaz:


README_FRAME_FORMAT.rst
```diff
@@ -62,7 +68,9 @@ Then it follows the info about the filter pipeline.  There is place for a pipeli
       |   +--number of filters
       +--[msgpack] fixext 16
 
-In addition, a frame can be completed with meta-information about the stored data; these data blocks are called metalayers and it is up to the user to store whatever data they want there, with the only (strong) suggestion that they have to be in the msgpack format.  Here it is the format for the case that there exist some metalayers::
+At the end of the frame, meta-information about the chunked data can be stored in data blocks are called metalayers.
```
FrancescAlted, line 71: I suggest the next rephrasing:

At the end of the frame come the so-called metalayers, that is, meta-information about the data stored in the frame.

README_FRAME_FORMAT.rst
```diff
@@ -164,23 +146,33 @@ __________________________________________
...
 
 
-The trailer section
--------------------
+Trailer
+-------
 
-Here it is data that can change in size, mainly the `usermeta` chunk::
+The trailer for the frame is encoded via `msgpack <https://msgpack.org>`_ and contains a user meta data chunk and
+a fingerprint::
```
FrancescAlted, line 175: I would add a note here stating that the `usermeta` chunk can change in size during the life of the frame.  This is an important feature, and the reason why `usermeta` is in the trailer, and not in the header.

nmoinvaz:
I have made the requested changes, but also made a modification in the header regarding `size of map` which was unclear.

FrancescAlted:
Thanks @nmoinvaz !

PR #144, opened 2020-08-08
Added srcsize parameter to blosc2_decompress_ctx and blosc2_getitem_ctx.

nmoinvaz:
This PR adds additional checks to `src` buffers when passing into `blosc2_decompres_ctx` and `blosc2_getitem_ctx` to ensure that blosc2 library does not read past the end of the provided buffer. When reading it checks to see that there is sufficient space in `src` to be able to read the required bytes.

This should fix issues where the Blosc2 header might contain an corrupt/malicious size or offset value and try to read past the end of the `src` buffer.

### DIFF
--- a/bench/sum_openmp.c
+++ b/bench/sum_openmp.c
@@ -151,14 +151,14 @@ int main(void) {
       return 1;
     }
   }
-  cparams.compcode = codec;
+  cparams.compcode = (uint8_t)codec;
 
   long clevel = CLEVEL;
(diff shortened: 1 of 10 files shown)


FrancescAlted:
Ok.  I understand that this is critical for security, so let's merge it.  It supposes an API change, but it is not too bad (just one additional parameter in `blosc_decompress_ctx`), and we are in beta after all.  I'll add a note in the RELEASE_NOTES file afterwards.

PR #146, opened 2020-08-29
Add secure versions of blosc_compress and blosc_decompress

nmoinvaz:
This PR makes the following changes:

* Adds secure versions of `blosc_compress` and `blosc_decompress` for blosc2 called `blosc2_compress` and `blosc2_decompress`
* Cleans up parameter order in `blosc2_compress_ctx` to be similar to `blosc2_decompress_ctx` - should be `src` followed by `srcsize`.
  * Cleans up parameter order in `initialize_context_compression`  as well
* Changes fuzzers to use blosc2 secure versions of compress and decompress.

This PR should resolve OSS-fuzz issue: https://oss-fuzz.com/testcase-detail/4799687189331968.

```
Running 1 inputs
Running: /home/nathan/Source/c-blosc2/build/tests/fuzz/decompress_fuzzer /home/nathan/Downloads/clusterfuzz-testcase-decompress_fuzzer-4799687189331968
==124013==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x4ed820 in initialize_context_decompression /home/nathan/Source/c-blosc2/blosc/blosc2.c:1587:33
    #1 0x4ec4a5 in blosc_run_decompression_with_context /home/nathan/Source/c-blosc2/blosc/blosc2.c:2101:11
    #2 0x4ede48 in blosc2_decompress /home/nathan/Source/c-blosc2/blosc/blosc2.c:2192:12
    #3 0x4ee140 in blosc_decompress /home/nathan/Source/c-blosc2/blosc/blosc2.c:2203:10
    #4 0x49e8a3 in LLVMFuzzerTestOneInput /home/nathan/Source/c-blosc2/tests/fuzz/fuzz_decompress.c:33:5
    #5 0x49ee64 in main /home/nathan/Source/c-blosc2/tests/fuzz/standalone.c:32:7
    #6 0x7ffff6ee583f in __libc_start_main /build/glibc-e6zv40/glibc-2.23/csu/../csu/libc-start.c:291
    #7 0x4255c8 in _start (/home/nathan/Source/c-blosc2/build/tests/fuzz/decompress_fuzzer+0x4255c8)

  Uninitialized value was created by a heap allocation
    #0 0x4517ed in malloc /tmp/final/llvm.src/projects/compiler-rt/lib/msan/msan_interceptors.cc:916:3
    #1 0x49e842 in LLVMFuzzerTestOneInput /home/nathan/Source/c-blosc2/tests/fuzz/fuzz_decompress.c:31:12
    #2 0x49ee64 in main /home/nathan/Source/c-blosc2/tests/fuzz/standalone.c:32:7
    #3 0x7ffff6ee583f in __libc_start_main /build/glibc-e6zv40/glibc-2.23/csu/../csu/libc-start.c:291

SUMMARY: MemorySanitizer: use-of-uninitialized-value /home/nathan/Source/c-blosc2/blosc/blosc2.c:1587:33 in initialize_context_decompression
Exiting
```

### DIFF
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -964,7 +964,7 @@ int pipeline_d(blosc2_context* context, const int32_t bsize, uint8_t* dest,
 /* Decompress & unshuffle a single block */
 static int blosc_d(
     struct thread_context* thread_context, int32_t bsize,
-    int32_t leftoverblock, const uint8_t* src, size_t srcsize, uint8_t* dest,
+    int32_t leftoverblock, const uint8_t* src, int32_t srcsize, uint8_t* dest,
     int32_t offset, uint8_t* tmp, uint8_t* tmp2) {
   blosc2_context* context = thread_context->parent_context;
   uint8_t* filters = context->filters;
@@ -1373,7 +1373,7 @@ static void flags_to_filters(const uint8_t flags, uint8_t* filters) {
 
 
 static int initialize_context_compression(
-  blosc2_context* context, const void* src, size_t srcsize, void* dest,
+  blosc2_context* context, const void* src, int32_t srcsize, void* dest,
   int32_t destsize, int clevel, uint8_t const *filters,
   uint8_t const *filters_meta, int32_t typesize, int compressor,
   int32_t blocksize, int new_nthreads, int nthreads, blosc2_schunk* schunk) {
@@ -1814,8 +1814,8 @@ int blosc_compress_context(blosc2_context* context) {
 
 
 /* The public secure routine for compression with context. */
-int blosc2_compress_ctx(blosc2_context* context, const void* src, size_t srcsize,
-                        void* dest, size_t destsize) {
+int blosc2_compress_ctx(blosc2_context* context, const void* src, int32_t srcsize,
+                        void* dest, int32_t destsize) {
   int error, cbytes;
 
   if (context->do_compress != 1) {
@@ -1934,8 +1934,8 @@ void build_filters(const int doshuffle, const int delta,
 }
 
 /* The public secure routine for compression. */
-int blosc2_compress(int clevel, int doshuffle, size_t typesize,
-                    const void* src, size_t srcsize, void* dest, size_t destsize) {
+int blosc2_compress(int clevel, int doshuffle, int32_t typesize,
+                    const void* src, int32_t srcsize, void* dest, int32_t destsize) {
   int error;
   int result;
   char* envvar;
@@ -2077,12 +2077,12 @@ int blosc2_compress(int clevel, int doshuffle, size_t typesize,
 /* The public routine for compression. */
 int blosc_compress(int clevel, int doshuffle, size_t typesize, size_t nbytes,
                    const void* src, void* dest, size_t destsize) {
-  return blosc2_compress(clevel, doshuffle, typesize, src, nbytes, dest, destsize);
+  return blosc2_compress(clevel, doshuffle, (int32_t)typesize, src, (int32_t)nbytes, dest, (int32_t)destsize);
 }
 
 
-int blosc_run_decompression_with_context(blosc2_context* context, const void* src, size_t srcsize,
-                                         void* dest, size_t destsize) {
+int blosc_run_decompression_with_context(blosc2_context* context, const void* src, int32_t srcsize,
+                                         void* dest, int32_t destsize) {
   int32_t ntbytes;
   uint8_t* _src = (uint8_t*)src;
   uint8_t version;
@@ -2132,8 +2132,8 @@ int blosc_run_decompression_with_context(blosc2_context* context, const void* sr
 
 
 /* The public secure routine for decompression with context. */
-int blosc2_decompress_ctx(blosc2_context* context, const void* src, size_t srcsize,
-                          void* dest, size_t destsize) {
+int blosc2_decompress_ctx(blosc2_context* context, const void* src, int32_t srcsize,
+                          void* dest, int32_t destsize) {
   int result;
 
   if (context->do_compress != 0) {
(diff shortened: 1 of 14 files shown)


blosc/blosc2.c
```diff
@@ -1373,18 +1373,19 @@ static void flags_to_filters(const uint8_t flags, uint8_t* filters) {
 
 
 static int initialize_context_compression(
-  blosc2_context* context, int32_t sourcesize, const void* src, void* dest,
+  blosc2_context* context, const void* src, size_t srcsize, void* dest,
```
FrancescAlted, line 1376: You are using `size_t` here for `srcsize`, whereas `destsize` uses `int32_t`.  I am very much inclined to use explicit type sizes, so I'd suggest to converge both to `int32_t`.

blosc/blosc2.c
```diff
@@ -1812,9 +1813,9 @@ int blosc_compress_context(blosc2_context* context) {
 }
 
 
-/* The public routine for compression with context. */
-int blosc2_compress_ctx(blosc2_context* context, size_t nbytes,
-                        const void* src, void* dest, size_t destsize) {
+/* The public secure routine for compression with context. */
+int blosc2_compress_ctx(blosc2_context* context, const void* src, size_t srcsize,
+                        void* dest, size_t destsize) {
```
FrancescAlted, line 1818: To be consistent with the above, `srcsize` and `destsize` should be changed to `int32_t`.  Could you do it here?

blosc/blosc2.c
```diff
@@ -2124,7 +2131,7 @@ int blosc_run_decompression_with_context(blosc2_context* context, const void* sr
 }
 
 
-/* The public routine for decompression with context. */
+/* The public secure routine for decompression with context. */
 int blosc2_decompress_ctx(blosc2_context* context, const void* src, size_t srcsize,
                           void* dest, size_t destsize) {
```
FrancescAlted, line 2136: For consistency, `srcsize` and `destsize` types should be changed to `int32_t` too.

blosc/blosc2.h
```diff
@@ -271,48 +271,9 @@ BLOSC_EXPORT void blosc_destroy(void);
...
- *
- * **BLOSC_NOLOCK=(ANY VALUE)**: This will call *blosc2_compress_ctx()* under
- * the hood, with the *compressor*, *blocksize* and
- * *numinternalthreads* parameters set to the same as the last calls to
- * *blosc_set_compressor*, *blosc_set_blocksize* and
- * *blosc_set_nthreads*. *BLOSC_CLEVEL*, *BLOSC_SHUFFLE*, *BLOSC_DELTA* and
- * *BLOSC_TYPESIZE* environment vars will also be honored.
  *
+ * This function supports all the configuration based environment variables
+ * available in blosc2_compress.
```
FrancescAlted, line 276: +1 for removing redundancies.

blosc/blosc2.h
```diff
@@ -341,23 +302,9 @@ BLOSC_EXPORT int blosc_compress(int clevel, int doshuffle, size_t typesize,
...
- * **BLOSC_NTHREADS=(INTEGER)**: This will call
- * *blosc_set_nthreads(BLOSC_NTHREADS)* before the proper decompression
- * process starts.
- *
- * **BLOSC_NOLOCK=(ANY VALUE)**: This will call *blosc2_decompress_ctx*
- * under the hood, with the *numinternalthreads* parameter set to the
- * same value as the last call to *blosc_set_nthreads*.
  *
+ * This function supports all the configuration based environment variables
+ * available in blosc2_decompress.
```
FrancescAlted, line 307: Ditto.

FrancescAlted (changes requested):
This looks pretty good to me.  Try to address the type issue mentioned, and I think this is good to go.

blosc/blosc2.c
nmoinvaz: For some reason `get_coffsets` uses `int64_t` in `cbytes` and `off_cbytes`. Can these be changed to `int32_t`? It seems like size variables can be `int32_t` but position variables need to be `int64_t`..

blosc/blosc2.c
nmoinvaz: Yes I think fixed width is the best for an api.

blosc/blosc2.c
FrancescAlted: Yes, `get_coffsets` are designed to get offsets within frames, which are meant as 63-bit containers rather than 31.  I could have stored 31-bit chun lengths as the frame index, and then compute absolute offsets during reading, but I thought that storing 63-bit indexes would be easier (and compression would get rid of the additional storage).

nmoinvaz: The variables though are not representing offsets, it seems they are representing number of compressed bytes, so they can be `int32_t`?

nmoinvaz: Otherwise I am not sure why `cbytes` is allowed to be `int64_t` but yet we can only ever pass in `int32_t` into `blosc2_compress_ctx`. 

https://github.com/Blosc/c-blosc2/blob/f90c05489f5c5d4335f2a5e3338f57c9e0c13e4a/blosc/frame.c#L987

FrancescAlted: Sorry, I know this can be confusing for the newcomer.  The `cbytes` and `off_cbytes` from `get_coffsets` params refers to the length of the frame and offsets to each chunk inside the frame, so they are `int64_t` indeed.  For the sake of readibility, perhaps `cbytes` here should be renamed to something like `frame_len` or similar.  Hope this helps.

nmoinvaz: In the code in frame.c, We get the size of the `coffset_cbytes` which is `int64_t`. Then should we cast it to `int32_t` when calling `blosc2_decompress_ctx`?

```
  uint8_t* coffsets = get_coffsets(frame, header_len, cbytes, &coffsets_cbytes);
  ...
  int32_t off_nbytes = blosc2_decompress_ctx(dctx, coffsets, (int32_t)coffsets_cbytes, offsets, nchunks * 8);
```
Or should `srcsize` and `destsize` be `int64_t`?

FrancescAlted: After close inspection, `coffset_cbytes` is the amount of compressed bytes of the offsets chunk.  This number can never be larger than 2**31, so this is clearly an `int32_t`.  Sorry that I did not detected this when I merged your previous PR.  Could you fix it please?

FrancescAlted:
Looks good to me!

PR #148, opened 2020-09-13
Fixed wrong size being passed to blosc_run_decompression_with_context.

nmoinvaz:
This is related to #146. I checked the fuzzers this morning and the OSS-fuzz issue still remained and it looks like I missed a few things.


nmoinvaz:
I also ran this against https://oss-fuzz.com/testcase-detail/5120916286341120 and it appears to solve that fuzzer issue too. 2 birds 1 stone?

FrancescAlted:
Yep.  Well done.

PR #157, opened 2020-10-11
Fixed heap-buffer-overflow asan read error on 32-bit arch

nmoinvaz:
https://oss-fuzz.com/testcase-detail/6033831129513984

On 32-bit machines the `if (src + cbytes > src_end) {` check would be by-passed because `src + cbytes` would overflow.

It should also fix: 
https://oss-fuzz.com/testcase-detail/5095855033679872
https://oss-fuzz.com/testcase-detail/5736707137470464


FrancescAlted:
Is that ready for review?  If so, just put me in the reviewers list ;-)

FrancescAlted (approved):
Looks good!

PR #161, opened 2020-10-16
Fixed asan integer overflow when checking compressed size bounds.

nmoinvaz:
Sorry for the wack-a-mole.
https://oss-fuzz.com/testcase-detail/5146170315505664

### DIFF
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -1018,18 +1018,20 @@ static int blosc_d(
 
   neblock = bsize / nstreams;
   for (int j = 0; j < nstreams; j++) {
-    srcsize -= sizeof(int32_t);
-    if (srcsize < 0) {
+    if (srcsize < sizeof(int32_t)) {
       /* Not enough input to read compressed size */
       return -1;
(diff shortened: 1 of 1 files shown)


blosc/blosc2.c
```diff
@@ -1018,18 +1018,20 @@ static int blosc_d(
...
       return -1;
     }
+    srcsize -= sizeof(int32_t);
     cbytes = sw32_(src);      /* amount of compressed bytes */
-    src += sizeof(int32_t);
-    srcsize -= cbytes;
-    if (srcsize < 0) {
-      /* Not enough input to read compressed bytes */
-      return -1;
+    if (cbytes > 0) {
```
FrancescAlted, line 1027: Do we need to check for `cbytes <= 0` too?

nmoinvaz: That is handled further down in the code.

FrancescAlted (approved):

FrancescAlted:
Thanks @nmoinvaz !

PR #166, opened 2020-11-05
Added fuzzers for frame compression and decompression.

nmoinvaz:


nmoinvaz:
It appears there are segfaults because `assert` is used through the frame.c code. That probably needs to change..

FrancescAlted:
Yeah, those asserts were meant to help me during the development phase of the frame, but they should be not necessary anymore.  Can you develop a bit more why these are causing trouble?

nmoinvaz:
Yes, because when the frame fuzzers run with random data it will cause those areas to assert/segfault. Those areas will need to be handled safely and return some error code or NULL.

FrancescAlted:
> Yes, because when the frame fuzzers run with random data it will cause those areas to assert/segfault. Those areas will need to be handled safely and return some error code or NULL. Fuzzers are also run during ctest, and that is where it is failing in CI, because unexpected data is passed.

Do you need some additional input from me or you think you can manage this one yourself?

nmoinvaz:
No additional input is needed, thanks! I can do it when I get time again.

nmoinvaz:
I have changed all the `asserts`, but now CI fuzz is reporting fuzzing issues.

nmoinvaz:
Should be ready for review now that I have fixed all the fuzzing issues.

FrancescAlted (approved):
This looks awesome.  Thank you Nathan!

PR #170, opened 2020-11-17
First version (but tested!) of lazy chunks

FrancescAlted:
Lazy chunks are like regular chunk but without the block data initialized.  They just have all the metadata for allowing loading data on-demand.  This allows for better fine granularity when doing partial reads of data that is on-disk (via e.g. `blosc2_getitem_ctx`), as well as paralell read on disk (via `blosc2_decompress_ctx`).

One may think in using lazy chunks as a way to cache data that is on-disk via an array of booleans specifying whether a specific data block has been loaded or not.  However, this might be redundant with the OS page cache, so not sure is that this is worth implementing in the future.


nmoinvaz:
Looks interesting. I'm not sure if I am completely qualified to do a review, but I didn't see anything that stood out. The lazy-loading does seem to make the CHUNK format more complex. Is it backwards compatible? How are the lazy-loaded blocks loaded, it is asynchronously through a callback?

nmoinvaz (commented):
Looks interesting. I'm not sure if I am completely qualified to do a review, but I didn't see anything that stood out. The lazy-loading does seem to make the CHUNK format more complex. Is it backwards compatible? How are the lazy-loaded blocks loaded, it is asynchronously through a callback?

FrancescAlted:
Thanks for your comments @nmoinvaz .  Yeah, the lazy-loading works by looking into the lazy flag during decompression, and if true, the block is read on-the flight.  That means that chunks that are on-disk can be loaded and decompressed in parallel.  That can bring a good speed-up during reads.

The possible incompatibility of lazy chunks is a good point.  So as to isolate the generation of lazy chunks, I added a specific function for retrieving a lazy chunk (`blosc2_schunk_get_chunk_lazy`) and made it clear in docs that such a chunk can only be accessed via either `blosc2_decompress_ctx` (paralell) or `blosc2_getitem_ctx` (serial).

Unfortunately on my last commit I triggered some alarm in the fuzzer mechanism.  Thoughts?

nmoinvaz:
The fuzzing warning is interesting, because you don't appear to have changed that function in that last commit anyway.

nmoinvaz:
Here is a more readable version of the issue:
```
ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6030000016b0 at pc 0x00000052d917 bp 0x7fffe78f3a30 sp 0x7fffe78f31f8
READ of size 1 at 0x6030000016b0 thread T0
SCARINESS: 22 (1-byte-read-heap-buffer-overflow-far-from-bounds)
    #0 0x52d916 in __asan_memcpy /src/llvm-project/compiler-rt/lib/asan/asan_interceptors_memintrinsics.cpp:22:3
    #1 0x5ba23a in blosc_d /src/c-blosc2/blosc/blosc2.c:1029:5
    #2 0x5b0c9b in serial_blosc /src/c-blosc2/blosc/blosc2.c:1208:16
    #3 0x5b0c9b in do_job /src/c-blosc2/blosc/blosc2.c:1366:15
    #4 0x5b7319 in blosc_run_decompression_with_context /src/c-blosc2/blosc/blosc2.c:2148:13
    #5 0x5b7cdb in blosc2_decompress /src/c-blosc2/blosc/blosc2.c:2216:12
    #6 0x56136f in LLVMFuzzerTestOneInput /src/c-blosc2/tests/fuzz/fuzz_decompress.c:33:5
    #7 0x465e41 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:599:15
    #8 0x46534a in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:505:3
    #9 0x4676f4 in fuzzer::Fuzzer::ReadAndExecuteSeedCorpora(std::__Fuzzer::vector<fuzzer::SizedFile, fuzzer::fuzzer_allocator<fuzzer::SizedFile> >&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:814:7
    #10 0x467909 in fuzzer::Fuzzer::Loop(std::__Fuzzer::vector<fuzzer::SizedFile, fuzzer::fuzzer_allocator<fuzzer::SizedFile> >&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:845:3
    #11 0x455d7b in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:906:6
    #12 0x47fe22 in main /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerMain.cpp:20:10
    #13 0x7fee8dae783f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2083f)
    #14 0x42abd8 in _start (out/decompress_fuzzer+0x42abd8)

DEDUP_TOKEN: __asan_memcpy--blosc_d--serial_blosc
0x6030000016b0 is located 12 bytes to the right of 20-byte region [0x603000001690,0x6030000016a4)
allocated by thread T0 here:
    #0 0x52e56d in malloc /src/llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:145:3
    #1 0x443707 in operator new(unsigned long) (out/decompress_fuzzer+0x443707)
    #2 0x46534a in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:505:3
    #3 0x4676f4 in fuzzer::Fuzzer::ReadAndExecuteSeedCorpora(std::__Fuzzer::vector<fuzzer::SizedFile, fuzzer::fuzzer_allocator<fuzzer::SizedFile> >&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:814:7
    #4 0x467909 in fuzzer::Fuzzer::Loop(std::__Fuzzer::vector<fuzzer::SizedFile, fuzzer::fuzzer_allocator<fuzzer::SizedFile> >&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:845:3
    #5 0x455d7b in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:906:6
    #6 0x47fe22 in main /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerMain.cpp:20:10
    #7 0x7fee8dae783f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2083f)

DEDUP_TOKEN: malloc--operator new(unsigned long)--fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*)
```
And the output also always gives the base64 of the data that reproduced it:
```
AiACDQEAAAAPAAAAFAAAABMAbZ0=
```
This can be converted to hex and saved to a file and then that argument can be passed to `fuzz_decompress`.

blosc/blosc2.c
```diff
@@ -984,14 +984,52 @@ static int blosc_d(
...
+      return -13;
+    }
+  }
+
+  // If the chunk is memcpyed, we just have to copy the block to dest and return
+  int memcpyed = src[BLOSC2_CHUNK_FLAGS] & (uint8_t)BLOSC_MEMCPYED;
+  if (memcpyed) {
+    int32_t chunk_nbytes = *(int32_t*)(src + BLOSC2_CHUNK_NBYTES);
+    int bsize_ = leftoverblock ? chunk_nbytes % context->blocksize : bsize;
+    memcpy(dest + dest_offset, src + BLOSC_MAX_OVERHEAD + nblock * context->blocksize, bsize_);
```
nmoinvaz, line 1029: It looks like the fuzzer doesn't like this line.

blosc/blosc2.c
```diff
@@ -1164,27 +1202,12 @@ static int serial_blosc(struct thread_context* thread_context) {
       }
     }
     else {
-      if (memcpyed) {
-        // Check that sizes in header are compatible, otherwise there is a header corruption
-        int32_t csize = sw32_(context->src + 12);   /* compressed buffer size */
-        if (context->sourcesize + BLOSC_MAX_OVERHEAD != csize) {
```
nmoinvaz, line 1170: Perhaps it is missing two these equivalent checks.

FrancescAlted:
Thanks @nmoinvaz for the clues.  The problem with fuzzer seems to be fixed now.  I think I need to setup a README_FUZZER with instrucctions on how to proceed for debugging fuzzer issues.

README_FUZZER.md
```diff
@@ -0,0 +1,8 @@
+How to look into fuzzer issues
+==============================
+
+* Look into the output of the test case at: https://oss-fuzz.com/testcases?showall=1 (Nathan, I am not sure how to look into a specific testcase coming from the logs in CI; could you chime in and put the procedure here ?).  Use the blosc.oss.fuzz@gmail.com account so as to access it.
```
FrancescAlted, line 4: @nmoinvaz Could you please complete the testcase lookup procedure here?  Thanks in advance!

FrancescAlted:
Ok.  I think this is ready to go.  Lazychunks has more potential as chunk caches, but that will come in the future (if at all).

README_FUZZER.md
nmoinvaz: Oss-fuzz.com is only for people on the CC list in google/oss-fuzz project configuration yaml. Otherwise people can only view the `CI Fuzz` CI run output which is what I did in the case of this PR.

image
image

FrancescAlted: I see now.  I updated the `README_FUZZER.md` accordingly.  Thanks!

PR #175, opened 2020-12-21
Add BLOSC_TRACE macro

aleixalcacer:
Macro `BLOSC_TRACE_ERROR()` is implemented. It can be used to print messages to stderr when the env variable `BLOSC_TRACE` is defined.


FrancescAlted (changes requested):
Looks good.  It would be nice if you can add a section in the README file on how to activate error/warning messages.

PR #184, opened 2021-01-04
Sanity check for compressed and uncompressed sizes in frame header

nmoinvaz:
Should inadvertently resolve https://oss-fuzz.com/testcase-detail/5131542452109312.


FrancescAlted:
LGTM

PR #186, opened 2021-01-06
Use VERSION_GREATER to compare cmake versions

iris-ersc:
Using "GREATER" to compare cmake versions does not work as expected for version > 3.10.

See also: https://cmake.org/cmake/help/latest/variable/CMAKE_VERSION.html


FrancescAlted:
Thanks @iris-ersc !

PR #189, opened 2021-01-10
Fixes for arm64

FrancescAlted:
This PR adds support for Apple M1 arm64 architecture and fixes some bugs in the bitshuffle filter by not calling the NEON-optimized routines (looks like bit shuffling is failing for 1, 2 and 4 bytes type sizes).  At any rate, the clang 12 compiler with Mac OSX M1 does an excellent job at optimizing the scalar bitshuffle.

Also, the NEON shuffle routines have been reformatted and unrolled (also for readability).  These functions seems to be working fine, and they are indeed faster than its scalar counterparts, so the former are the ones that are actually used.

It would be nice to come with some way to test this with the arm64 arch on GHA.


blosc/blosc2.c
```diff
@@ -686,7 +686,7 @@ static bool get_run(const uint8_t* ip, const uint8_t* ip_bound) {
   memset(&value, x, 8);
   while (ip < (ip_bound - 8)) {
 #if defined(BLOSC_STRICT_ALIGN)
-    memcpy(&value2, ref, 8);
+    memcpy(&value2, ip, 8);
```
nmoinvaz: This looks like a previously undiscovered bug. Nice.

nmoinvaz:
I'm not an expert in intrinsics. Unrolling is a good idea. Formatting fixes makes it easier to read.

FrancescAlted:
Ok.  Thanks for having look @nmoinvaz .  Will merge and will add testing on ARM when GHA would support them.

PR #191, opened 2021-01-12
Fixed ASAN error when reading blosc version in blosc_getitem.

nmoinvaz:
It seems we need to check to see if header length exceeds frame length which will stop the error from even happening.
  https://oss-fuzz.com/testcase-detail/6522073996460032


FrancescAlted:
LGTM.

PR #192, opened 2021-01-12
fixed memory leaks

ihsinme:
the functions has the ability to exit without first clearing the allocated memory.

I propose to fix this error.


FrancescAlted:
Thanks for providing this @ihsinme !  The fuzzer tests are crashing though.  @nmoinvaz any hints here?

nmoinvaz:
Yes, in `frame_get_metalayers`, the variable `header` should only be freed if `(frame->sdata == NULL)`, otherwise it ends up in a double-free of `frame->sdata`.

ihsinme:
sorry for the inattention.
indeed, memory for header is not allocated in some cases.
I've corrected.

ihsinme:
sorry for this rubbish.
correcting the PR conflict, I most likely attracted many new mistakes into it.

FrancescAlted:
LGTM

PR #195, opened 2021-01-14
Zero runlen

FrancescAlted:
In this PR:

* Optimized containers for chunks with all zero, all NaNs and all equal values

* Optimized offsets for frames when chunks are all zeros or all NaNs (not equal values).

* Bonus: Added Blosc1 compatibility tests.


FrancescAlted:
Not quite ready for merging (more tests should be done, specially for more backends than just in-memory frames), but I'd appreciate a review.  In particular, fuzzing is not happy with this yet...

blosc/schunk.c
```diff
@@ -314,7 +314,37 @@ int blosc2_schunk_append_chunk(blosc2_schunk *schunk, uint8_t *chunk, bool copy)
...
+    else if (all_zeros || all_nans) {
+      if (schunk->cbytes == 0) {
+        // Increase by 0 just to be explicit
+        schunk->cbytes += 0;
+      }
+    }
+    else {
+      // A non-special chunk is always stored in frames
+      schunk->cbytes += cbytes;
+    }
```
aleixalcacer, line 338: This can be simplified:
```
if (all_zeros || all_nans) {
  schunk->cbytes += 0;
} else {
  schunk->cbytes += cbytes;
}
```

examples/CMakeLists.txt
```diff
@@ -1,12 +1,17 @@
 # sources for examples
 set(SOURCES contexts delta_schunk_ex multithread simple frame_metalayers noinit find_roots)
 if(NOT DEACTIVATE_LZ4)
-    set(SOURCES ${SOURCES} schunk_simple frame_simple frame_big frame_backed_schunk compress_file eframe_simple)
+    set(SOURCES ${SOURCES} schunk_simple frame_simple frame_backed_schunk compress_file)
```
aleixalcacer, line 4: Why have `eframe_simple` and `frame_big` files been removed?

blosc/schunk.c
aleixalcacer, line 338: This can be simplified:
```
if (all_zeros || all_nans) {
  schunk->cbytes += 0;
}
else {
  schunk->cbytes += cbytes;
}
```

examples/CMakeLists.txt
```diff
@@ -1,12 +1,17 @@
 # sources for examples
 set(SOURCES contexts delta_schunk_ex multithread simple frame_metalayers noinit find_roots)
 if(NOT DEACTIVATE_LZ4)
-    set(SOURCES ${SOURCES} schunk_simple frame_simple frame_big frame_backed_schunk compress_file eframe_simple)
```
aleixalcacer, line 4: Why has `eframe_simple` been deleted?

aleixalcacer (changes requested):

blosc/schunk.c
FrancescAlted: Not possible because `all_repeats` includes `all_zeros` and `all_nans` (both bits are set at the same time).

examples/CMakeLists.txt
FrancescAlted: Fixed in 428dcf6.

blosc/frame.c
```diff
@@ -462,7 +463,7 @@ int get_header_info(blosc2_frame *frame, int32_t *header_len, int64_t *frame_len
     }
 
     // Sanity check for compressed sizes
-    if ((*cbytes < 0) || (*nbytes > 0 && *cbytes == 0) || ((int64_t)*nchunks * *chunksize < *nbytes)) {
+    if ((*cbytes < 0) || ((int64_t)*nchunks * *chunksize < *nbytes)) {
```
FrancescAlted, line 466: With the support of zero and NaN run-lengths as special values in frame offsets, `(*nbytes > 0 && *cbytes == 0)` is perfectly valid.  I suppose this is the reason why fuzzer tests are complaining.  Possibly the the frames in `tests/fuzz/corpus` needs to be redone because the format of frames has changed. @nmoinvaz ?

aleixalcacer (approved):

PR #197, opened 2021-01-24
Fixed ASAN error due to invalid header length.

nmoinvaz:
https://oss-fuzz.com/testcase-detail/6021380745330688
In this instance header length ended up being less than zero.


FrancescAlted:
LGTM

PR #198, opened 2021-01-24
Fixed UBSAN error when offset exceeds header length.

nmoinvaz:
https://oss-fuzz.com/testcase-detail/4800450727641088


FrancescAlted:
LGTM

PR #199, opened 2021-01-24
Fixed UBSAN divide-by-zero error in handle_runlen with typesize.

nmoinvaz:
https://oss-fuzz.com/testcase-detail/6044163185442816


FrancescAlted:
LGTM

PR #200, opened 2021-01-24
Prevent ASAN 64-bit memory allocation in fuzzers.

nmoinvaz:
https://oss-fuzz.com/testcase-detail/5005175230169088
Also see https://github.com/google/sanitizers/issues/802.


FrancescAlted:
LGTM.

PR #202, opened 2021-01-26
Fixed ASAN heap-buffer-overflow by 4 when reading usermeta length.

nmoinvaz:
https://oss-fuzz.com/testcase-detail/6308459295408128


FrancescAlted:
LGTM

PR #203, opened 2021-01-26
Fixed segfault on unknown address when not checking offsets.

nmoinvaz:
https://oss-fuzz.com/testcase-detail/6501619711737856


FrancescAlted:
LGTM

PR #204, opened 2021-01-27
Fixed frame metalayer loading memory leaks

nmoinvaz:
https://oss-fuzz.com/testcase-detail/4783158937649152

There are a few commits, you might want to review them one by one. 

https://github.com/Blosc/c-blosc2/pull/204/commits/888435d0697ff7516ea5cb6f48c7648a1b3df848 changes it so that all `schunk` is freed through `blosc2_schunk_free`, this helps on failure, so that the entire object is freed properly. I have also added a `avoid_frame_free` to `schunk` similar to how `frame` has `avoid_sdata_free` which helps to know when `schunk` should free `frame` or not.

https://github.com/Blosc/c-blosc2/pull/204/commits/c0a8138ff991ba192612c154f4e61e520cabbe2a is improved in that it checks against the `header_len` and not the `frame_len` which prevents invalid data earlier when `header_len` is too small.

### DIFF
--- a/blosc/frame.c
+++ b/blosc/frame.c
@@ -282,6 +282,9 @@ void *new_header_frame(blosc2_schunk *schunk, blosc2_frame *frame) {
 
   // Now, deal with metalayers
   int16_t nmetalayers = schunk->nmetalayers;
+  if (nmetalayers < 0 || nmetalayers > BLOSC2_MAX_METALAYERS) {
+    return NULL;
+  }
 
   // Make space for the header of metalayers (array marker, size, map of offsets)
   h2 = realloc(h2, (size_t)hsize + 1 + 1 + 2 + 1 + 2);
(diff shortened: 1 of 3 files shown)


nmoinvaz:
This should also resolve https://oss-fuzz.com/testcase-detail/5985680524181504.

blosc/frame.c
```diff
@@ -1139,102 +1142,47 @@ int32_t frame_get_usermeta(blosc2_frame* frame, uint8_t** usermeta) {
...
-      fclose(fp);
-    }
-    if (rbytes != (size_t) header_len) {
-      BLOSC_TRACE_ERROR("Cannot access the header out of the fileframe.");
-      free(header);
-      return -2;
-    }
-  }
+static int frame_get_metalayers_from_header(blosc2_frame* frame, blosc2_schunk* schunk, uint8_t* header,
+  int32_t header_len) {
```
FrancescAlted, line 1146: Watch out indentation. Our style (undocumented) is to align parameters with previous line.

FrancescAlted (changes requested):
Well done!  I have just a small comment on coding style, but other than this, that's an excellent PR.

FrancescAlted:
Curiously enough Win32 complains here too.  Maybe somehow related to #205 ?

FrancescAlted (changes requested):
I am fine with this as long as win32 issues are addressed.

FrancescAlted (approved):

FrancescAlted:
Merged as the Win32 issues are unrelated to this.

PR #205, opened 2021-01-27
Fixed ASAN heap-buffer-overflow when reading frame index chunk.

nmoinvaz:
https://oss-fuzz.com/testcase-detail/5963546544439296

@FrancescAlted the `get_coffsets` function is confusing to me. I'm not sure what part it is reading from the frame. I _assume_ that it is reading from the `chunks` section mention in the [FRAME format documentation](https://github.com/Blosc/c-blosc2/blob/master/README_FRAME_FORMAT.rst). And that in the `chunks` section it is trying to read the Blosc chunk called `chunk idx`. Please correct me if I am wrong. The variable names in the section are kind of opaque.

I am assuming:
- `header_len` = frame header length.
- `cbytes` = compressed data size? What exactly is this referring to and why would we seek to this? Does this not include the size of the `chunk idx`?
- I know it is reading into a Blosc chunk because I see it is reading a 32-bit integer at `BLOSC2_CHUNK_CBYTES` offset.

Thanks!

### DIFF
--- a/blosc/frame.c
+++ b/blosc/frame.c
@@ -914,17 +914,15 @@ uint8_t* get_coffsets(blosc2_frame *frame, int32_t header_len, int64_t cbytes, i
     return frame->coffsets;
   }
   if (frame->sdata != NULL) {
-    if (header_len + cbytes > frame->len) {
+    int32_t off_pos = header_len + cbytes;
+    // Check that there is enough room to read Blosc header
+    if (off_pos + BLOSC_EXTENDED_HEADER_LENGTH > frame->len) {
       BLOSC_TRACE_ERROR("Cannot read the offsets past frame boundary.");
       return NULL;
     }
     // For in-memory frames, the coffset is just one pointer away
-    uint8_t* off_start = frame->sdata + header_len + cbytes;
+    uint8_t* off_start = frame->sdata + off_pos;
     if (off_cbytes != NULL) {
-      if (header_len + cbytes + BLOSC2_CHUNK_CBYTES + (signed)sizeof(int32_t) > frame->len) {
-        BLOSC_TRACE_ERROR("Cannot read the offsets compressed size past frame boundary.");
-        return NULL;
-      }
       *off_cbytes = *(int32_t*) (off_start + BLOSC2_CHUNK_CBYTES);
     }
     return off_start;
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -2673,6 +2673,11 @@ int blosc2_getitem_ctx(blosc2_context* context, const void* src, int32_t srcsize
   uint8_t* _src = (uint8_t*)(src);
   int result;
 
+  if (srcsize < BLOSC_MIN_HEADER_LENGTH) {
+    /* Not enough input to parse Blosc1 header */
+    return -1;
+  }
+
   /* Minimally populate the context */
   context->typesize = _src[BLOSC2_CHUNK_TYPESIZE];
   context->blocksize = sw32_(_src + BLOSC2_CHUNK_BLOCKSIZE);


FrancescAlted:
> https://oss-fuzz.com/testcase-detail/5963546544439296
> 
> @FrancescAlted the `get_coffsets` function is confusing to me.

Yeah, you are confused because the naming is confusing.  The reason is mainly historical.  At the beginning, we started by implementing the continuous frame, and in the chunk idx we were storing offsets to the different chunks in the frame.  However, this changed when we decided to implement the 'extended' frame (BTW, we are not yet sure about its naming; it could probably change to 'sparse' frame or 'distributed' frame; or we may be even call it just 'frame' and change the original one to 'contiguous' frame).  In extended frames, the chunk idx section stores chunk IDs (essentially its sequential number as was appended) and not offsets (hence why the name is not correct anymore).

So we need a more thorough naming change in the code.  My proposal would be: `offset` -> `chunk_id` and essentially get rid of the chunk offset name except for those cases where it makes a explicit reference to a contiguous frame

> I'm not sure what part it is reading from the frame. I _assume_ that it is reading from the `chunks` section mention in the [FRAME format documentation](https://github.com/Blosc/c-blosc2/blob/master/README_FRAME_FORMAT.rst). And that in the `chunks` section it is trying to read the Blosc chunk called `chunk idx`. Please correct me if I am wrong. The variable names in the section are kind of opaque.

You are correct here.  Sorry for the confusion.

> 
> I am assuming:
> 
> * `header_len` = frame header length.

Correct.

> * `cbytes` = compressed data size? What exactly is this referring to and why would we seek to this? Does this not include the size of the `chunk idx`?

`cbytes` is the amount of bytes that the actual compressed data chunks takes in the frame, and it does *not* include chunk idx. 

> * I know it is reading into a Blosc chunk because I see it is reading a 32-bit integer at `BLOSC2_CHUNK_CBYTES` offset.

Yes.  The chunk idx is easy to locate using a `header_len + cbytes` offset.  And yes, the chunk idx is a regular Blosc2 chunk.

Now that I think, maybe it would make more sense to actually have 4 sections in the frame: header, chunks, chunk idx and trailer.
Agreed?

> Thanks!

Rather the contrary, thanks to your efforts in understanding the code because they are not only making Blosc2 safer and cleaner, but specially much better documented!

nmoinvaz:
> Now that I think, maybe it would make more sense to actually have 4 sections in the frame: header, chunks, chunk idx and trailer.
Agreed?

Yes. That would make more sense to me.

> cbytes is the amount of bytes that the actual compressed data chunks takes in the frame, and it does not include chunk idx.

So `cbytes` is the compressed size of `chunks` section? Aren't each chunk compressed already? Does the `chunks` section get recompressed in a frame?

>  In extended frames, the chunk idx section stores chunk IDs (essentially its sequential number as was appended) and not offsets (hence why the name is not correct anymore).

So even in normal frames, it is no longer an offset?

> So we need a more thorough naming change in the code. My proposal would be: offset -> chunk_id and essentially get rid of the chunk offset name except for those cases where it makes a explicit reference to a contiguous frame

Yes maybe `chunk_id` or `chunk_order` would be good. It should be determined if they represent an ID or an sequential order and then document it. Because we don't want people to assume it can be used for ordering if it is just a unique ID.

nmoinvaz:
I think according to your comments, this PR is still ok. Can you rerun the failed test?

FrancescAlted:
> > Now that I think, maybe it would make more sense to actually have 4 sections in the frame: header, chunks, chunk idx and trailer.
> > Agreed?
> 
> Yes. That would make more sense to me.
> 
> > cbytes is the amount of bytes that the actual compressed data chunks takes in the frame, and it does not include chunk idx.
> 
> So `cbytes` is the compressed size of `chunks` section? Aren't each chunk compressed already? Does the `chunks` section get recompressed in a frame?

Yes, `cbytes` is the compressed size of `chunks` section.  Regarding about the 'recompression' question, well, a frame is just a different way to store a super-chunk.  You can even create a super-chunk using a frame as storage (instead of the default in-memory format).  Remember that a frame is just a format, and it can be either on-disk or in-memory.  The new extended frame (this is probably going to be renamed as 'disperse' frame) is another format (more closely tied to the default in-memory format).  So you don't actually need to re-compress anything so as to create frames: they are just another storage backend.

> 
> > In extended frames, the chunk idx section stores chunk IDs (essentially its sequential number as was appended) and not offsets (hence why the name is not correct anymore).
> 
> So even in normal frames, it is no longer an offset?

Yes, in normal frames the chunk ID is the offset (it is the best key for quickly finding a chunk in the frame).

> 
> > So we need a more thorough naming change in the code. My proposal would be: offset -> chunk_id and essentially get rid of the chunk offset name except for those cases where it makes a explicit reference to a contiguous frame
> 
> Yes maybe `chunk_id` or `chunk_order` would be good. It should be determined if they represent an ID or an sequential order and then document it. Because we don't want people to assume it can be used for ordering if it is just a unique ID.

That's a good point.  As we are supporting insertions and updates, I think `chunk_id` would be a better choice, but we need to ponder a bit more about this.

blosc/frame.c
```diff
@@ -914,17 +914,15 @@ uint8_t* get_coffsets(blosc2_frame *frame, int32_t header_len, int64_t cbytes, i
     return frame->coffsets;
   }
   if (frame->sdata != NULL) {
-    if (header_len + cbytes > frame->len) {
+    int32_t off_pos = header_len + cbytes;
+    // Check that there is enough room to read Blosc header
+    if (off_pos + BLOSC_MIN_HEADER_LENGTH > frame->len) {
```
FrancescAlted, line 919: You should replace `BLOSC_MIN_HEADER_LENGTH` by `BLOSC_EXTENDED_HEADER_LENGTH`, as frames can only host extended (Blosc2) frames.

FrancescAlted (approved):
Looks good.  I am only suggesting a small nit pick.

blosc/frame.c
nmoinvaz: Ok that should be fixed now.

FrancescAlted:
Looks like Win32 keeps complaining here.

FrancescAlted:
Definitively this seems a general regression in the Win32 arch (my latest commits on updating lz44 and zstd codecs are failing there too).  Heck, is GHA offering a trusting platform for Win32?  Besides, I am not sure whether we need Win32 support; probably not.  @nmoinvaz thoughts?

nmoinvaz:
I don't think the problem is with GHA. I have often noticed sometimes Win32 tests randomly failing. I think it is a bug. I ran all the tests on my Windows machine, building for 32-bit, but am unable to reproduce the segfault. I develop primarily on Windows but for this project it is mainly Ubuntu. I don't think we should get rid of it Win32 GHA. If you have a Windows machine maybe you can try to reproduce it.

nmoinvaz:
I have used Windows GHA CI instances for `zlib-ng` and `minizip-ng` projects. It is not like Travis which has diminished Windows offerings. GHA's Windows offering is the real deal.

FrancescAlted:
> I don't think the problem is with GHA. I have often noticed sometimes Win32 tests randomly failing. I think it is a bug. I ran all the tests on my Windows machine, building for 32-bit, but am unable to reproduce the segfault. I develop primarily on Windows but for this project it is mainly Ubuntu. I don't think we should get rid of it Win32 GHA. If you have a Windows machine maybe you can try to reproduce it.

Fair enough.  I am not an expert on Windows, but now that Win32 is failing more consistently perhaps it is a good time to have a deeper look into this.

FrancescAlted:
Merging, as the Win32 issues are unrelated to this.

FrancescAlted:
For what is worth, valgrind is not reporting nothing suspicious on the both examples that occasionally crashes on Win32:

```
faltet@corex14 ~/b/c/build (master)> valgrind ./examples/frame_backed_schunk                                                                       (base)
==18880== Memcheck, a memory error detector
==18880== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==18880== Using Valgrind-3.16.1 and LibVEX; rerun with -h for copyright info
==18880== Command: ./examples/frame_backed_schunk
==18880==
Blosc version info: 2.0.0.beta.6.dev ($Date:: 2020-04-21 #$)
Compression ratio: 381.5 MB -> 3.9 MB (98.2x)
Time for append data to a schunk backed by an in-memory frame: 4.87 s, 78.3 MB/s
Compression ratio: 381.5 MB -> 3.9 MB (98.2x)
Time for append data to a schunk backed by a fileframe: 4.83 s, 79.0 MB/s
Successful roundtrip data <-> schunk (frame-backed) !
==18880==
==18880== HEAP SUMMARY:
==18880==     in use at exit: 2,791 bytes in 8 blocks
==18880==   total heap usage: 4,304 allocs, 4,296 frees, 1,629,958,921 bytes allocated
==18880==
==18880== LEAK SUMMARY:
==18880==    definitely lost: 0 bytes in 0 blocks
==18880==    indirectly lost: 0 bytes in 0 blocks
==18880==      possibly lost: 0 bytes in 0 blocks
==18880==    still reachable: 2,791 bytes in 8 blocks
==18880==         suppressed: 0 bytes in 0 blocks
==18880== Rerun with --leak-check=full to see details of leaked memory
==18880==
==18880== For lists of detected and suppressed errors, rerun with: -s
==18880== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
faltet@corex14 ~/b/c/build (master)> valgrind ./examples/zstd_dict                                                                                 (base)
==18930== Memcheck, a memory error detector
==18930== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==18930== Using Valgrind-3.16.1 and LibVEX; rerun with -h for copyright info
==18930== Command: ./examples/zstd_dict
==18930==
Blosc version info: 2.0.0.beta.6.dev ($Date:: 2020-04-21 #$)
Compression ratio: 30.5 MB -> 1.0 MB (30.9x)
Compression time: 22.5 s, 1.4 MB/s
Decompression time: 0.492 s, 62.0 MB/s
Successful roundtrip!
==18930==
==18930== HEAP SUMMARY:
==18930==     in use at exit: 2,791 bytes in 8 blocks
==18930==   total heap usage: 1,666 allocs, 1,658 frees, 1,065,569,454 bytes allocated
==18930==
==18930== LEAK SUMMARY:
==18930==    definitely lost: 0 bytes in 0 blocks
==18930==    indirectly lost: 0 bytes in 0 blocks
==18930==      possibly lost: 0 bytes in 0 blocks
==18930==    still reachable: 2,791 bytes in 8 blocks
==18930==         suppressed: 0 bytes in 0 blocks
==18930== Rerun with --leak-check=full to see details of leaked memory
==18930==
==18930== For lists of detected and suppressed errors, rerun with: -s
==18930== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

Heck, debugging these kind of eigen-bugs without the help of valgrind is going to be tough.

PR #206, opened 2021-01-28
Lazychunk2

FrancescAlted:
This PR removes the need for allocating the compressed data blocks in lazy chunks, making the process of reading from persistent storage less memory demanding.


nmoinvaz:
I can make edits to the chunk doc in a separate PR after this one is merged if you want.

FrancescAlted:
> I can make edits to the chunk doc in a separate PR after this one is merged if you want.

Sounds good!

PR #207, opened 2021-01-29
Rename Extended-Frame to Disperse-Frame

martaiborra:
Change name of Extended-Frame to Disperse-Frame in order to be more explicit.

### DIFF
--- a/README_DFRAME_FORMAT.rst
+++ b/README_DFRAME_FORMAT.rst
@@ -1,8 +1,8 @@
-Blosc2 Disperse-Frame Format
-============================
+Blosc2 Dispersed-Frame Format
+=============================
 
-Blosc (as of version 2.0.0) has an disperse-frame format that allows for the storage of different Blosc data chunks
-disperse, on-disk.
+Blosc (as of version 2.0.0) has a dispersed-frame format that allows for the storage of different Blosc data chunks
+dispersed, on-disk.
 
 When creating a dframe one must denote the `storage.sequential` as false and provide a name (which will be a directory)
 in `storage.urlpath` for the dframe to be stored. It is recommended to name the directory with the `.b2dframe`
--- a/README_DFRAME_FORMAT.rst
+++ b/README_DFRAME_FORMAT.rst
@@ -33,7 +33,7 @@ Examples
 
 Structure example
 ^^^^^^^^^^^^^^^^^
-As shown below, an dframe of 4 chunks will be composed of a directory with each chunk file and the frame file::
+As shown below, a dframe of 4 chunks will be composed of a directory with each chunk file and the frame file::
 
  dir.b2dframe/
  │
--- a/README_DFRAME_FORMAT.rst
+++ b/README_DFRAME_FORMAT.rst
@@ -1,14 +1,14 @@
-Blosc2 Extended-Frame Format
+Blosc2 Disperse-Frame Format
 ============================
 
-Blosc (as of version 2.0.0) has an extended-frame format that allows for the storage of different Blosc data chunks
-sparse, on-disk.
+Blosc (as of version 2.0.0) has an disperse-frame format that allows for the storage of different Blosc data chunks
+disperse, on-disk.
 
-When creating an eframe one must denote the `storage.sequential` as false and provide a name (which will be a directory)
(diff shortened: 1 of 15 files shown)


nmoinvaz:
I think perhaps it should be "sparse".
image

image
image

README_DFRAME_FORMAT.rst
```diff
@@ -33,9 +33,9 @@ Examples
 
 Structure example
 ^^^^^^^^^^^^^^^^^
-As shown below, an eframe of 4 chunks will be composed of a directory with each chunk file and the frame file::
+As shown below, an dframe of 4 chunks will be composed of a directory with each chunk file and the frame file::
```
FrancescAlted, line 36: A dframe

blosc/dframe.c
```diff
@@ -37,7 +37,7 @@
 
 
 /* Append an existing chunk into an extended frame. */
```
FrancescAlted, line 39: extended - > disperse

blosc/dframe.c
```diff
@@ -56,7 +56,7 @@ void* eframe_create_chunk(blosc2_frame* frame, uint8_t* chunk, int32_t nchunk, i
 
 
 /*Get chunk from extended frame. */
```
FrancescAlted, line 58: Ditto.

Perhaps you can do a (supervised) automatic replace for extended - > disperse in the code too.

FrancescAlted (changes requested):

FrancescAlted:
> I think perhaps it should be "sparse".
> image
> Instead of "disperse".
> image
> Sparse is adjective and disperse is verb form.

Ups, we did not pay enough attention to disperse being a verb. Yeah, sparse was our initial choice, but then the meaning 2 (meager) put us off. But perhaps we should only pay attention to meaning 1 and go ahead with sparse. After all @nmoinvaz is a native speaker and he finds it a better choice anyways. @martaiborra @aleixalcacer , thoughts?

FrancescAlted:
On the other hand, if we adopt sparse, we have a name collision with the short name for the 'serialized' frame (`sframe`) already existing in the code (and we need to find yet another name :-). 'distributed' frame can also be a possibility, but it gives me a sense of distribution among different machines. Oh, the joy of giving names, isn't it fun?

README_DFRAME_FORMAT.rst
martaiborra: Fixed in 2426e5ae2a82ab94f3fc05902aeb22df78beb0b3 .

blosc/dframe.c
martaiborra: Fixed in 2426e5ae2a82ab94f3fc05902aeb22df78beb0b3 .

blosc/dframe.c
martaiborra: Fix in 2426e5ae2a82ab94f3fc05902aeb22df78beb0b3 .

nmoinvaz:
I dont' understand the use of the term _serialized_ in the context of a memory-backed frame. I would think that saying it is serialized doesn't add any value (because all frames use msgpack - if you are considering the use of msgpack as the serialization mechanism).

I think it might be better to remove the use of the existing term _sframe_ which is used for _serialized frame_. Instead functions like `blosc2_frame_from_sframe` can become `blosc2_frame_from_memory` or `blosc2_frame_from_buffer`. It would more nicely go along with `blosc2_frame_from_file`. Use of the variable `sdata` in `blosc2_frame` can just become `data` or `buffer`.

FrancescAlted:
Thanks @nmoinvaz  for your thoughts.  I agree that we need a simplification in the different naming for frames.  I am not even sure that frames need to be public (or at least we should not need to disclose the internal details).  I'll think more about this and will come with a proposal (probably in the form of PR).

@martaiborra until we accept the new proposal, I think we need to freeze this PR.

FrancescAlted:
We discussed this internally, and we decided to go with 'sparse' frames.  Also, we are gonna try to get rid of all references to frames in the API; specifying `blosc2_storage` should be the only thing a user needs to know about.  Also, the user can always go to the frame description documents in case she wants to know more about the internal storage format.

Closing.

PR #208, opened 2021-01-30
Fixed UBSAN divide-by-zero and check for invalid blocksize and typesize.

nmoinvaz:
https://oss-fuzz.com/testcase-detail/6011475527335936
https://oss-fuzz.com/testcase-detail/4559366210584576

### DIFF
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -1565,9 +1565,18 @@ static int initialize_context_compression(
   }
 
   /* Check typesize limits */
+  if (context->typesize <= 0) {
+    BLOSC_TRACE_ERROR("`typesize` is less than or equal to zero.");
+    return -1;
+  }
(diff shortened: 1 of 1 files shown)


blosc/blosc2.c
```diff
@@ -1565,10 +1565,15 @@ static int initialize_context_compression(
   }
 
   /* Check typesize limits */
-  if (context->typesize > BLOSC_MAX_TYPESIZE) {
-    /* If typesize is too large, treat buffer as an 1-byte stream. */
+  if (context->typesize <= 0 || context->typesize > BLOSC_MAX_TYPESIZE) {
```
FrancescAlted, line 1568: I would consider a negative typesize as an error.

nmoinvaz: Fixed.

FrancescAlted (approved):

PR #209, opened 2021-01-30
Fixed ASAN unknown-read when reading stream 

nmoinvaz:
https://oss-fuzz.com/testcase-detail/4727058852872192

At the bottom of the loop `src += cbytes`. If `cbytes` is not set to 0 then it will seek backwards. Same for `nbytes`.


FrancescAlted (approved):
Well spotted

PR #210, opened 2021-01-30
Fixed ASAN direct-leak of frame in blosc2_frame_to_schunk.

nmoinvaz:
https://oss-fuzz.com/testcase-detail/5854816091897856

This was semi-introduced in my previous direct-leak fix.

I believe this logic is correct - when `blosc2_frame_to_schunk` has `copy = true` then it will copy internally, and it doesn't need a reference to the frame hence `schunk->frame = NULL`. But if `copy = false` then it will need to keep around (and assume ownership of the frame) in otherwords, `schunk->frame = frame`.  So when `copy = true` we can `avoid_frame_free`, but when `copy = false` we assume frame ownership and must free.

Can you check that to make sure I got it right? Or maybe there is a better way to handle this. The function that gets called is `blosc2_schunk_open_sframe`. It creates a `frame` and then uses that to create a `schunk`. It can either keep the `frame` and not copy, or it can free the `frame` and copy.

### DIFF
--- a/blosc/frame.c
+++ b/blosc/frame.c
@@ -1307,7 +1307,7 @@ blosc2_schunk* blosc2_frame_to_schunk(blosc2_frame* frame, bool copy) {
 
   blosc2_schunk* schunk = calloc(1, sizeof(blosc2_schunk));
   schunk->frame = frame;
-  schunk->avoid_frame_free = !copy;
+  schunk->avoid_frame_free = copy;
   int ret = get_header_info(frame, &header_len, &frame_len, &schunk->nbytes, &schunk->cbytes,
                             &schunk->chunksize, &schunk->nchunks, &schunk->typesize,
                             &schunk->compcode, &schunk->clevel, schunk->filters, schunk->filters_meta);
--- a/blosc/schunk.c
+++ b/blosc/schunk.c
@@ -293,10 +293,8 @@ blosc2_schunk* blosc2_schunk_open_sframe(uint8_t *sframe, int64_t len) {
   if (frame == NULL) {
     return NULL;
   }
+  // Super-chunk will assume ownership of frame and free it
   blosc2_schunk* schunk = blosc2_frame_to_schunk(frame, false);
-  if (schunk == NULL) {
-    blosc2_frame_free(frame);
-  }
   return schunk;
 }
 


FrancescAlted:
> https://oss-fuzz.com/testcase-detail/5854816091897856
> 
> This was semi-introduced in my previous direct-leak fix.
> 
> I believe this logic is correct - when `blosc2_frame_to_schunk` has `copy = true` then it will copy internally, and it doesn't need a reference to the frame hence `schunk->frame = NULL`. 

Correct.

> But if `copy = false` then it will need to keep around (and assume ownership of the frame) in otherwords, `schunk->frame = frame`. So when `copy = true` we can `avoid_frame_free`, but when `copy = false` we assume frame ownership and must free.

Not quite. My intention here is that in blosc2_frame_to_schunk, the schunk should never take frame ownership. Transferring the ownership is a possibility, but I don't think this would be a good idea (unless you can convince me of the contrary). 
> 
> Can you check that to make sure I got it right? Or maybe there is a better way to handle this. The function that gets called is `blosc2_schunk_open_sframe`. It creates a `frame` and then uses that to create a `schunk`. It can either keep the `frame` and not copy, or it can free the `frame` and copy.

nmoinvaz:
> Not quite. My intention here is that in blosc2_frame_to_schunk, the schunk should never take frame ownership. Transferring the ownership is a possibility, but I don't think this would be a good idea (unless you can convince me of the contrary).

I just mean to say that when `copy = false`, then `schunk` becomes responsible for freeing the frame. 

As an alternative, we can remove the code that free's the frame from `blosc2_schunk_free`. If there is no transfer of owner then it doesn't make since to have it freeing the frame. Instead `blosc2_schunk_open_sframe` can just call `blosc2_frame_to_schunk` to copy the frame and afterward we can remove it. This might require changes to other parts of code.

FrancescAlted:
As expressed in a previous comment, we discussed this internally and we decided:

1) Add a new `blosc2_schunk_copy()` that we can use for copying super-chunks, and convert to a different storage format, if needed.

2) Remove all the APIs that make reference to frames, and add a couple like `blosc2_schunk_to_buffer` and `blosc2_schunk_from_buffer`.

With that, there shouldn't be a need to 'transfer' ownership: the super-chunk will always be responsible to release its storage.  So, let's put this PR on-hold until we manage to implement 1) and 2).  After that, this will probably go away.

FrancescAlted:
On a second thought, I am going to merge this because when the frame would disappear from the public API, the transfer of ownership won't be a problem.  Hopefully, this will fix the outstanding leak report by fuzzer.

PR #213, opened 2021-01-30
Fixed MSAN use-of-uninitialized in ZSTD_checkContinuity. 

nmoinvaz:
In this instance `bsize = 1` and `nstream > 1` (while debugging was 2) resulting in `neblock` being 0 and `zstd_wrap_decompress` being called with `maxout` of 0. 

Maybe there is some other better check I could also be doing?

https://oss-fuzz.com/testcase-detail/6243129091031040

### DIFF
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -1115,6 +1115,10 @@ static int blosc_d(
   }
 
   neblock = bsize / nstreams;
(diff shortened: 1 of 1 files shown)


FrancescAlted (approved):
Yes, I think this is a proper check.

PR #215, opened 2021-01-31
Fixed MSAN use-of-uninitialized in tinfl_decompress when invalid dist.

nmoinvaz:
In this instance `dist` was 31 which `s_dist_base` translates as 0.

https://github.com/Blosc/c-blosc2/blob/452c3a9d2ba548718a590245f88c217dc2a99b88/internal-complibs/miniz-2.0.8/miniz.c#L2655-L2657

https://oss-fuzz.com/testcase-detail/4863557237473280


FrancescAlted (approved):
Great.  Although it would be cool to replace mini by zlib-ng (as soon as it is considered stable).

nmoinvaz:
This fix has been merged upstream too!

PR #217, opened 2021-01-31
Added custom BLOSC2 error codes.

nmoinvaz:


blosc/blosc2.h
```diff
@@ -250,6 +250,34 @@ enum {
     BLOSC2_VALUE_RUNLEN = 0x3,    //!< generic value run-length
 };
 
+/**
+ * @brief Error codes
+ */
+enum {
+  BLOSC2_ERROR_SUCCESS = 0,           //<! Success
```
FrancescAlted, line 257: Maybe it would be better to rename `BLOSC2_ERROR_SUCCESS`-> `BLOSC2_SUCCESS` ?

FrancescAlted (approved):
Excellent patch!  This work was long due, thank you!

I have a simple suggestion, but we can change it later (if you agree).

blosc/blosc2.h
nmoinvaz: I could go either way. If you want to change it I'm ok or we can change it in the future.

PR #222, opened 2021-02-04
Cleanup blosc header reading and writing

nmoinvaz:
This PR cleans up the way blosc min and extended header are read and written by the library. 

- In blosc2.c, I have cleaned up the run-length code as well as the new header code simplifies things. 
- In blosc2.c, `blosc2_getitem_ctx` has a new security parameter called `destsize`.
- In CMake, a change from `blosc_shared_testing` has been changed to `blosc_testing` and it has been made static so there is no need to copy it. I have been experiencing some issues on Windows, where I would recompile and it would not allow me to break into blosc library and I believe it had to do with this - in some instances not copying the dll.
- In CMake, I fixed a few instances where `lib` PREFIX was not added on Windows. 
- There is a new `BLOSC2_RUNLEN_MASK` which allows for masking out values of the `blosc2_flags` that are run-length related: `(header->blosc2_flags >> 4) & BLOSC2_RUNLEN_MASK`

### DIFF
--- a/bench/CMakeLists.txt
+++ b/bench/CMakeLists.txt
@@ -33,26 +33,6 @@ target_link_libraries(trunc_prec_schunk blosc_shared_testing)
 target_link_libraries(sum_openmp blosc_shared_testing)
 target_link_libraries(zero_runlen blosc_shared_testing)
 
-
-# have to copy blosc dlls on Windows
-if(MSVC)
-    add_custom_command(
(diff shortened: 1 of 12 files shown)


FrancescAlted (approved):
This looks really good to me!  +10 for merging this.

I am just a bit worried that this could collide with the big private frame PR (#223).  But I guess we should not know if we don't try.  So let's go!

PR #223, opened 2021-02-04
Private frames

FrancescAlted:
This is a big refactoring tackling the next points:

* The `blosc2_frame` struct is not public anymore.

* The public APIs for dealing with frames are gone.  Now frames are just a storage option.  This should make for a good relief in the cognitive load of users.

* As a consequence, a lot of renaming has happened internally (and affected some public functions).  In particular, the old `sframe` has been renamed to `framebuf` and `eframe` -> `sframe` i.e. the sparse frame name, as suggested by @nmoinvaz  is now there (I like sparse too).


README_SFRAME_FORMAT.rst
```diff
@@ -1,15 +1,14 @@
-Blosc2 Extended-Frame Format
-============================
+Blosc2 Sparse Frame Format
+==========================
 
-Blosc (as of version 2.0.0) has an extended-frame format that allows for the storage of different Blosc data chunks
-sparse, on-disk.
+Blosc (as of version 2.0.0) has a sparse frame (sframe for short) format that allows for the storage of different Blosc data chunks on disk in a sparse way.
```
nmoinvaz, line 4: Consider:

... allows for the _non-contiguous_ storage of ~~different~~ Blosc data chunks on disk.

README_SFRAME_FORMAT.rst
```diff
@@ -1,15 +1,14 @@
...
 
-When creating an eframe one must denote the `storage.sequential` as false and provide a name (which will be a directory)
-in `storage.urlpath` for the eframe to be stored. It is recommended to name the directory with the `.b2eframe`
+When creating an sparse frame one must denote the `storage.sequential` as false and provide a name (which will be a directory)
+in `storage.urlpath` for the sframe to be stored. It is recommended to name the directory with the `.b2sframe`
 extension.
 
-An eframe is made up of a frame file and the chunks stored in the same directory on-disk.
-The frame follows the format described in the `frame format <README_FRAME_FORMAT.rst>`_ document, with the difference
+An sframe is made up of a frame file and the chunks stored in the same directory on-disk.
```
nmoinvaz, line 10: Consider:

... is made up of _multiple files: a frame index chunk and individually stored data chunks in the same directory on-disk_. 

In this way:
frame index chunk (or index chunk for short as also referred to it further in the document)
data chunks = as distinguished from frame index chunk.

blosc/sframe.h
```diff
@@ -6,11 +6,11 @@
...
 
-#ifndef BLOSC_EFRAME_H
-#define BLOSC_EFRAME_H
+#ifndef BLOSC_SFRAME_H
+#define BLOSC_SFRAME_H
 
 
-void* eframe_create_chunk(blosc2_frame* frame, uint8_t* chunk, int32_t nchunk, int64_t cbytes);
-int eframe_get_chunk(blosc2_frame* frame, int32_t nchunk, uint8_t** chunk, bool* needs_free);
+void* sframe_create_chunk(blosc2_frame_s* frame, uint8_t* chunk, int32_t nchunk, int64_t cbytes);
```
nmoinvaz, line 13: Are these functions part of the public API?

blosc/schunk.c
```diff
@@ -293,41 +279,108 @@ blosc2_schunk* blosc2_schunk_copy(blosc2_schunk *schunk, blosc2_storage storage)
...
   }
   return new_schunk;
 }
 
 
 /* Open an existing super-chunk that is on-disk (no copy is made). */
-blosc2_schunk* blosc2_schunk_open(const blosc2_storage storage) {
-  if (storage.urlpath == NULL) {
-    BLOSC_TRACE_ERROR("You need to supply a storage.urlpath.");
+blosc2_schunk* blosc2_schunk_open(char* urlpath) {
```
nmoinvaz, line 293: Should be `const char *`?

blosc/blosc2.c
```diff
@@ -1258,10 +1261,10 @@ static int blosc_d(
     int32_t block_csize = block_csizes[nblock];
     // Read the lazy block on disk
     FILE* fp = NULL;
-    if (context->schunk->frame->eframe) {
+    if (frame->sframe) {
       // The chunk is not in the frame
-      char* chunkpath = malloc(strlen(context->schunk->frame->urlpath) + 1 + 8 + strlen(".chunk") + 1);
-      sprintf(chunkpath, "%s/%08X.chunk", context->schunk->frame->urlpath, nchunk);
+      char* chunkpath = malloc(strlen(frame->urlpath) + 1 + 8 + strlen(".chunk") + 1);
```
nmoinvaz, line 1266: I think there should be a helper function that opens chunk file, something like:
```
static inline FILE *frame_open_chunk_file(const char *urlpath, int32_t nchunk) {
  char* chunkpath = malloc(strlen(urlpath) + 1 + 8 + strlen(".chunk") + 1);
  sprintf(chunkpath, "%s/%08X.chunk", urlpath, nchunk);
  fp = fopen(chunkpath, "rb");
  free(chunkpath);
  return fp;
}
```
I noticed many places (in frame.c in general / not related to this PR) where `sprintf` is used in this manner and it could be encapsulated into helper functions.

blosc/schunk.c
```diff
@@ -377,14 +430,17 @@ int blosc2_schunk_free(blosc2_schunk *schunk) {
...
+  blosc2_frame_s* frame = frame_from_framebuf(framebuf, len, false);
   if (frame == NULL) {
     return NULL;
   }
-  // Super-chunk will assume ownership of frame and free it
-  blosc2_schunk* schunk = blosc2_frame_to_schunk(frame, false);
+  blosc2_schunk* schunk = frame_to_schunk(frame, copy);
+  if (copy) {
+    // We don't need the frame anymore
+    free(frame);
```
nmoinvaz, line 442: Should we use `frame_free` in case frame is partially initialized?

blosc/schunk.c
```diff
@@ -377,14 +430,17 @@ int blosc2_schunk_free(blosc2_schunk *schunk) {
 }
 
 
-/* Create a super-chunk out of a serialized frame (no copy is made). */
-blosc2_schunk* blosc2_schunk_open_sframe(uint8_t *sframe, int64_t len) {
-  blosc2_frame* frame = blosc2_frame_from_sframe(sframe, len, false);
+/* Create a super-chunk out of a frame buffer */
+blosc2_schunk* blosc2_schunk_from_buffer(uint8_t *framebuf, int64_t len, bool copy) {
+  blosc2_frame_s* frame = frame_from_framebuf(framebuf, len, false);
```
nmoinvaz, line 435: Maybe rename to `frame_from_buffer`?

blosc/schunk.c
```diff
@@ -293,41 +279,108 @@ blosc2_schunk* blosc2_schunk_copy(blosc2_schunk *schunk, blosc2_storage storage)
...
+int64_t frame_to_file(blosc2_frame_s* frame, char* urlpath) {
+  FILE* fp = fopen(urlpath, "wb");
+  size_t nitems = fwrite(frame->framebuf, (size_t)frame->len, 1, fp);
+  fclose(fp);
+  return nitems * (size_t)frame->len;
+}
+
+
+/* Write super-chunk out to a file. */
+int64_t blosc2_schunk_to_file(blosc2_schunk* schunk, char* urlpath) {
```
nmoinvaz, line 354: Should be `const char *urlpath`?

blosc/schunk.c
```diff
@@ -293,41 +279,108 @@ blosc2_schunk* blosc2_schunk_copy(blosc2_schunk *schunk, blosc2_storage storage)
...
+    blosc2_schunk_free(schunk_copy);
+  }
+
+  return framebuf_len;
+
+}
+
+
+/* Write an in-memory frame out to a file. */
+int64_t frame_to_file(blosc2_frame_s* frame, char* urlpath) {
```
nmoinvaz, line 345: Should be `const char *urlpath`?

blosc/schunk.c
```diff
@@ -293,41 +279,108 @@ blosc2_schunk* blosc2_schunk_copy(blosc2_schunk *schunk, blosc2_storage storage)
...
+  fclose(fp);
+  return nitems * (size_t)frame->len;
+}
+
+
+/* Write super-chunk out to a file. */
+int64_t blosc2_schunk_to_file(blosc2_schunk* schunk, char* urlpath) {
+  if (urlpath == NULL) {
+    BLOSC_TRACE_ERROR("urlpath cannot be NULL");
+    return -1;
```
nmoinvaz, line 357: Use `BLOSC2_ERROR_INVALID_PARAM` instead of -1.

blosc/frame.c
```diff
@@ -383,24 +383,24 @@ void *new_header_frame(blosc2_schunk *schunk, blosc2_frame *frame) {
...
     size_t rbytes = 0;
     FILE* fp = NULL;
-    if (frame->eframe) {
-      char* eframe_name = malloc(strlen(frame->urlpath) + strlen("/chunks.b2frame") + 1);
-      sprintf(eframe_name, "%s/chunks.b2frame", frame->urlpath);
-      fp = fopen(eframe_name, "rb");
-      free(eframe_name);
+    if (frame->sframe) {
+      char* sframe_name = malloc(strlen(frame->urlpath) + strlen("/chunks.b2frame") + 1);
+      sprintf(sframe_name, "%s/chunks.b2frame", frame->urlpath);
```
nmoinvaz, line 401: Consider putting these into helper function, see my other comment.

blosc/frame.c
```diff
@@ -1024,13 +980,13 @@ int frame_update_header(blosc2_frame* frame, blosc2_schunk* schunk, bool new) {
   }
 
   FILE* fp = NULL;
-  if (frame->sdata == NULL) {
+  if (frame->framebuf == NULL) {
```
nmoinvaz, line 983: Why not use `buffer` instead of `framebuf`?

nmoinvaz (commented):
Overall good changes. I made a few suggestions. 

I would also recommend squashing/combining the "WIP" comments into one. This can be done with most Git GUI or rebase tool. It will make the Git history cleaner too. 

On a side-note also consider still changing the merge type for the GitHub project to only "Rebase and Merge" to prevent merge commits I think is a good idea.

image

Prevents this:

image

blosc/sframe.h
FrancescAlted: No, they are not prefixed with `BLOSC_EXPORT`, so no public.

README_SFRAME_FORMAT.rst
FrancescAlted: Yes, I like the (non-)contiguous qualifier.  Perhaps I'd change `storage->sequential` to `storage->contiguous`.  Agreed?

blosc/schunk.c
FrancescAlted: Ok.  It *was* `const char *` before, but CLion was issuing warnings down in the inner code and this was my easiest fix.  I can revisit that again, although I don't think this is that important.

blosc/blosc2.c
FrancescAlted: Ok.  Happy to see suggestions for helpers.

blosc/frame.c
FrancescAlted: I was hesitating and decided for `framebuf`.  I think `framebuf` is more explicit about the contents of the buffer, but I don't have a strong opinion on this.

FrancescAlted:
> Overall good changes. I made a few suggestions. Sorry in advance if it is too much.
> 
> I would also recommend squashing/combining the "WIP" commits into one. This can be done with most Git GUI or rebase tool. It will make the Git history cleaner too.
> 
> On a side-note also consider still changing the merge type for the GitHub project to only "Rebase and Merge" to prevent merge commits I think is a good idea.
> 
> image
> 
> Prevents this:
> 
> image

Ok.  I have been too lazy to understand the implications of rebase vs plain merge, so thanks for the advise.  I have removed the possibility of "Merge pull request" in the GitHub configuration, so in the future we should do a rebase (or "squash and merge") before merging.  Sorry for the mess.

blosc/schunk.c
FrancescAlted: Fixed

blosc/schunk.c
FrancescAlted: Indeed!

blosc/schunk.c
FrancescAlted: Fixed.

blosc/schunk.c
FrancescAlted: Fixed.

blosc/schunk.c
FrancescAlted: Fixed.

blosc/schunk.c
FrancescAlted: Fixed.

blosc/schunk.c
FrancescAlted: I think I like using `cframe` for contiguous frames.  I did the change already.

blosc/schunk.c
FrancescAlted: Fixed

blosc/schunk.c
FrancescAlted: Fixed

blosc/frame.c
FrancescAlted: I ended using `cframe`, which is more meaningful IMO.

README_SFRAME_FORMAT.rst
FrancescAlted: I went ahead and did the change.  It looks better now :-)

nmoinvaz:
Changes so far look good. I am wondering why there are two difffernt extensions .cframe and .sframe. Why not just have both be named .frame. And the frame header will contain what type it is. Maybe that is how it already works?

README_SFRAME_FORMAT.rst
FrancescAlted: done.

FrancescAlted:
Yes, the idea of adding metainfo in the frame header about whether this is a cframe or sframe is good.  I think I am going to add it soon.  Provided this, right now we are only discriminating that we are opening a sframe if the urlpath is a directory, but if we add this property as metainfo, then we don't necessarily depend on the dir/file distinction.

This also opens the door to drop the `.b2cframe` and `.b2sframe` extensions in favor of just, say, `.b2frame` (or `.b2f` for short).

Finally we had some internal discussions about the way to store sframes.  The way it is now is:

```
dir.b2sframe/
 │
 ├── 00000000.chunk
 │
 ├── 00000001.chunk
 │
 ├── 00000002.chunk
 │
 ├── 00000003.chunk
 │
 └── chunks.b2frame
```

But if we put the cframe/sframe metainfo in the frame header we could also do:

```
sframe_name.b2frame  (meta and index data)
sframe_name/
 │
 ├── 00000000.chunk
 │
 ├── 00000001.chunk
 │
 ├── 00000002.chunk
 │
 ├── 00000003.chunk
```

The latter would also allow to handle a sframe exactly the same way than a cframe, with the only difference that chunks are accessed differently (in a local directory or in the future maybe on a key/value store or even in the network).  Thoughts?

nmoinvaz:
The way the zip format handles it is it stores the disk number the file can be found on. So you only need to check if disk number != current disk then switch disks and read content. But I am in favor of having a single frame extension tells the program where to find the chunks.

You might be able to do some bit tricks with the offset. If it is some way then

FrancescAlted:
> Changes so far look good. I am wondering why there are two difffernt extensions .cframe and .sframe. Why not just have both be named .frame. And the frame header will contain what type it is. Maybe that is how it already works?

I implemented exactly this in cdb3c92.  After that I think this should be ready for merging.

FrancescAlted:
In the end I decided to keep using a directory for storing both the chuns and the index inside it for our persistent sparse implementation; this allows for a more self-contained data inside a sparse frame.  When/if we support other backends than filesystems for storage we can leverage the new meta info about the frame type in the frame header (see cdb3c92).

FrancescAlted:
I am merging this.  @nmoinvaz your suggestion for helpers for `sprintf` et al. is still pending.  Feel free to address this on a future PR.  Thanks the the useful review!

nmoinvaz:
I have tested latest master and this PR looks like it may fix a number of OSS-Fuzz issues. Will be able to tell in a day or so.

PR #226, opened 2021-02-08
Allow to store a map of metainfo in trailer

aleixalcacer:
This allows to store different (a map of them actually) variable-length meta information in the trailer section.  This complements the capability to store fixed-length metainfo in the header.  As the trailer is re-written in every chunk operation (append, update, insert, removal), it is recommended that large and fixed size metainfo go to the header, and leave the rest to the new variable-length metainfo section in trailer.


FrancescAlted (changes requested):
Great job!  I have some small suggestions.

Regarding the fuzzer tests that are failing, I think that we would need to rebuild the `tests/fuzz/corpus/`.  @nmoinvaz can probably shed some light there.

nmoinvaz:
Yes, the frame corpus probably does need to be updated at some point. But it isn't the cause of any issues. The corpus is just used by the fuzzers as a starting point. The CI Fuzz gives the base64 encoded file that produces the error and also a stack trace if you want to take a look at it. It says a wild read so perhaps we are not checking `srcsize` somewhere.

blosc/schunk.c
```diff
@@ -1007,57 +1016,148 @@ int blosc2_get_metalayer(blosc2_schunk *schunk, const char *name, uint8_t **cont
...
   }
+  rc = frame_update_header(schunk->frame, schunk, false);
+  if (rc < 0) {
+    BLOSC_TRACE_ERROR("Unable to update metalayers into frame.");
+    return rc;
+  }
+  rc = frame_update_trailer(schunk->frame, schunk);
+  if (rc < 0) {
+    BLOSC_TRACE_ERROR("Unable to update trailer into frame.");
+    return rc;
```
nmoinvaz, line 1050: Don't need to `return rc`.

blosc/schunk.c
```diff
@@ -1007,57 +1016,148 @@ int blosc2_get_metalayer(blosc2_schunk *schunk, const char *name, uint8_t **cont
...
-    return usermeta_cbytes;
+int vlmetalayer_flush(blosc2_schunk* schunk) {
+  int rc;
+  if (schunk->frame == NULL) {
+    return BLOSC2_ERROR_SUCCESS;
   }
+  rc = frame_update_header(schunk->frame, schunk, false);
+  if (rc < 0) {
+    BLOSC_TRACE_ERROR("Unable to update metalayers into frame.");
+    return rc;
```
nmoinvaz, line 1045: Don't need to `return rc`.

blosc/schunk.c
```diff
@@ -1007,57 +1016,148 @@ int blosc2_get_metalayer(blosc2_schunk *schunk, const char *name, uint8_t **cont
...
-  void* usermeta_chunk = malloc(content_len + BLOSC_MAX_OVERHEAD);
-  blosc2_context *cctx = blosc2_create_cctx(cparams);
-  int usermeta_cbytes = blosc2_compress_ctx(cctx, content, content_len, usermeta_chunk,
-                                            content_len + BLOSC_MAX_OVERHEAD);
-  blosc2_free_ctx(cctx);
-  if (usermeta_cbytes < 0) {
-    free(usermeta_chunk);
-    return usermeta_cbytes;
+int vlmetalayer_flush(blosc2_schunk* schunk) {
+  int rc;
```
nmoinvaz, line 1038: Initialize `rc` to zero.

blosc/frame.c
```diff
@@ -1243,6 +1310,169 @@ int frame_get_metalayers(blosc2_frame_s* frame, blosc2_schunk* schunk) {
...
+                                               int32_t trailer_len) {
+  int64_t trailer_pos = FRAME_TRAILER_VLMETALAYERS;
+
+  // Get the size for the index of metalayers
+  uint16_t idx_size;
+  trailer_pos += 2 + sizeof(idx_size);
+  if (trailer_len < trailer_pos) {
+    return BLOSC2_ERROR_READ_BUFFER;
+  }
+  big_store(&idx_size, trailer + FRAME_TRAILER_VLMETALAYERS + 2, sizeof(idx_size));
```
nmoinvaz, line 1323: These few lines might be the cause of the fuzzer issue. You should ensure that the `trailer + FRAME_TRAILER_VLMETALAYERS + 4` does not exceed the length of the frame.

aleixalcacer: I have downloaded the artifact that CI Fuzz has uploaded. I understand that the `decompress_frame_fuzzer-testcase` is the file that produces the error and its contains a cframe plus it length at the end. Am I right? 

After some debugging using this file, I could not find anything wrong. I think the problem is in the file creation, but I'm not sure (the `msgpack2json` command line tool does not parse correctly the header of the `decompress_frame_fuzzer-testcase` file).

Anyway, this part of the code has been improved assuring that the length is checked. It would be nice if you can shed more light of the above error. Thanks!

FrancescAlted (approved):
Looks good to me.  Hope @nmoinvaz can help debugging the issues in fuzzer tests.

PR #227, opened 2021-02-09
Fixed UBSAN error due to invalid return value from frame_get_usermeta

nmoinvaz:
https://oss-fuzz.com/testcase-detail/5701192787296256


FrancescAlted:
LGTM

PR #228, opened 2021-02-09
Fixed ASAN potential leak in frame_special_chunk during error.

nmoinvaz:
https://oss-fuzz.com/testcase-detail/4936127276515328


FrancescAlted:
LGTM

PR #230, opened 2021-02-09
Fixed ASAN negative-size-param when reading content_len in frame_get_metalayers.

nmoinvaz:
https://oss-fuzz.com/testcase-detail/5964915892420608


nmoinvaz:
This should also fix:
https://oss-fuzz.com/testcase-detail/6121485435535360

FrancescAlted (approved):

PR #231, opened 2021-02-09
Fixed ASAN timeout when reading negative special offset.

nmoinvaz:
https://oss-fuzz.com/testcase-detail/6468303474262016


FrancescAlted (approved):
Great fix!

PR #233, opened 2021-02-14
Fixed ASAN heap-buffer-overflow from read trailer outside of frame.

nmoinvaz:
https://oss-fuzz.com/testcase-detail/5143072323272704


nmoinvaz:
It seems `frame_update_trailer` fails on `trailer_offset + trailer_len > frame->len` which in `test_copy` is `127 > 117`. Off-by-ten. So this could be a bug.

nmoinvaz:
Got it fixed now.

FrancescAlted (changes requested):
Looks good. I just have a couple of suggestions.

aleixalcacer:
The requested changes are already fixed. I'm going to merge the pull.

PR #237, opened 2021-02-17
Implement bench sframe and use lazychunks when possible

martaiborra:
- Implemented a bench to compare Sparse frames with Contiguous frames when updating, reordering or inserting chunks.
- Fixed bug when using lazychunks and changing the offsets in a Sparse frame.
- Use lazychunks when possible.


martaiborra:
Already done as separated pull requests.

PR #240, opened 2021-02-17
Fixed UBSAN unknown read when content_len < 0 in frame_get_vlmetalayers.

nmoinvaz:
https://oss-fuzz.com/testcase-detail/4622970490322944

This is one instance where I missed from #230.


FrancescAlted:
LGTM.

PR #243, opened 2021-02-18
Serialize blocksize into frame header

aleixalcacer:


FrancescAlted (commented):
It would be nice if the new blocksize metadata is used during e.g. super-chunk copies.  It maybe worth to use this PR so as to do this.

PR #244, opened 2021-02-19
Fixed UBSAN unknown read when trying to access invalid off_cbytes.

nmoinvaz:
- Added `blosc2_getitem` function which adds to `blosc_getitem` params `srcsize` and `destsize` to make sure we don't read past frame boundary, or write past destination buffer.

https://oss-fuzz.com/testcase-detail/5507944642904064.

### DIFF
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -1209,8 +1209,8 @@ static int blosc_d(
   int32_t compformat = (context->header_flags & (uint8_t)0xe0) >> 5u;
   int dont_split = (context->header_flags & (uint8_t)0x10) >> 4u;
   int memcpyed = context->header_flags & (uint8_t)BLOSC_MEMCPYED;
-  int32_t chunk_nbytes = *(int32_t*)(src + BLOSC2_CHUNK_NBYTES);
-  int32_t chunk_cbytes = *(int32_t*)(src + BLOSC2_CHUNK_CBYTES);
+  int32_t chunk_nbytes = sw32_(src + BLOSC2_CHUNK_NBYTES);
+  int32_t chunk_cbytes = sw32_(src + BLOSC2_CHUNK_CBYTES);
   //uint8_t blosc_version_format = src[BLOSC2_CHUNK_VERSION];
   int nstreams;
   int32_t neblock;
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -2553,10 +2553,7 @@ int _blosc_getitem(blosc2_context* context, blosc_header* header, const void* sr
   return ntbytes;
 }
 
-
-/* Specific routine optimized for decompression a small number of
-   items out of a compressed chunk.  Public non-contextual API. */
-int blosc_getitem(const void* src, int start, int nitems, void* dest) {
+int blosc2_getitem(const void* src, int32_t srcsize, int start, int nitems, void* dest, int32_t destsize) {
   blosc2_context context;
   int result;
 
@@ -2567,13 +2564,19 @@ int blosc_getitem(const void* src, int start, int nitems, void* dest) {
   context.nthreads = 1;  // force a serial decompression; fixes #95
 
   /* Call the actual getitem function */
-  result = blosc2_getitem_ctx(&context, src, INT32_MAX, start, nitems, dest, INT32_MAX);
+  result = blosc2_getitem_ctx(&context, src, srcsize, start, nitems, dest, destsize);
 
   /* Release resources */
   if (context.serial_context != NULL) {
     free_thread_context(context.serial_context);
   }
   return result;
+}
+
+/* Specific routine optimized for decompression a small number of
+   items out of a compressed chunk.  Public non-contextual API. */
+int blosc_getitem(const void* src, int start, int nitems, void* dest) {
+  return blosc2_getitem(src, INT32_MAX, start, nitems, dest, INT32_MAX);
 }
 
 int blosc2_getitem_ctx(blosc2_context* context, const void* src, int32_t srcsize,
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -1209,9 +1209,8 @@ static int blosc_d(
   int32_t compformat = (context->header_flags & (uint8_t)0xe0) >> 5u;
   int dont_split = (context->header_flags & (uint8_t)0x10) >> 4u;
   int memcpyed = context->header_flags & (uint8_t)BLOSC_MEMCPYED;
-  int32_t chunk_nbytes = sw32_(src + BLOSC2_CHUNK_NBYTES);
-  int32_t chunk_cbytes = sw32_(src + BLOSC2_CHUNK_CBYTES);
-  //uint8_t blosc_version_format = src[BLOSC2_CHUNK_VERSION];
+  int32_t chunk_nbytes;
+  int32_t chunk_cbytes;
   int nstreams;
   int32_t neblock;
   int32_t nbytes;                /* number of decompressed bytes in split */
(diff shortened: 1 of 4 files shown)


blosc/blosc2.c
```diff
@@ -1211,8 +1211,8 @@ static int blosc_d(
   int32_t compformat = (context->header_flags & (uint8_t)0xe0) >> 5u;
   int dont_split = (context->header_flags & (uint8_t)0x10) >> 4u;
   int memcpyed = context->header_flags & (uint8_t)BLOSC_MEMCPYED;
-  int32_t chunk_nbytes = *(int32_t*)(src + BLOSC2_CHUNK_NBYTES);
-  int32_t chunk_cbytes = *(int32_t*)(src + BLOSC2_CHUNK_CBYTES);
+  int32_t chunk_nbytes = sw32_(src + BLOSC2_CHUNK_NBYTES);
+  int32_t chunk_cbytes = sw32_(src + BLOSC2_CHUNK_CBYTES);
```
nmoinvaz, line 1215: I would honestly more prefer to re-read the blosc header using `blosc_read_header`. It has more checks to it. Thoughts?

blosc/frame.c
```diff
@@ -1025,7 +1025,7 @@ int64_t frame_from_schunk(blosc2_schunk *schunk, blosc2_frame_s *frame) {
 uint8_t* get_coffsets(blosc2_frame_s *frame, int32_t header_len, int64_t cbytes, int32_t *off_cbytes) {
   if (frame->coffsets != NULL) {
     if (off_cbytes != NULL) {
-      *off_cbytes = *(int32_t *) (frame->coffsets + BLOSC2_CHUNK_CBYTES);
+      *off_cbytes = sw32_(frame->coffsets + BLOSC2_CHUNK_CBYTES);
```
nmoinvaz, line 1028: Also prefer to re-read the entire blosc header here too.

blosc/frame.c
```diff
@@ -1043,7 +1043,7 @@ uint8_t* get_coffsets(blosc2_frame_s *frame, int32_t header_len, int64_t cbytes,
     // For in-memory frames, the coffset is just one pointer away
     uint8_t* off_start = frame->cframe + off_pos;
     if (off_cbytes != NULL) {
-      *off_cbytes = *(int32_t*) (off_start + BLOSC2_CHUNK_CBYTES);
+      *off_cbytes = sw32_(off_start + BLOSC2_CHUNK_CBYTES);
```
nmoinvaz, line 1046: Also prefer to re-read the entire blosc header here too.

blosc/blosc2.c
FrancescAlted: OMG, that bug was bad.  Curiously nobody never reported about this, which is probably a testimony that there is not a lot of users (well, none :-) creating chunks in one machine and decompressing them on a big-endian box.  However, this could be changing a lot for Blosc2 because frames makes data transport much more easier.

So +1 on using a `blosc_read_header` every time one need to do that.

blosc/blosc2.h
```diff
@@ -441,6 +441,24 @@ BLOSC_EXPORT int blosc2_chunk_repeatval(size_t nbytes, size_t typesize,
...
+ * @brief Get @p nitems (of @p typesize size) in @p src buffer starting in @p start.
+ * The items are returned in @p dest buffer. The dest buffer should have enough space
+ * for storing all items. This function is a more secure version of #blosc_getitem.
+ *
+ * @param src The compressed buffer from data will be decompressed.
+ * @param srcsize Size of the compressed buffer from data will be decompressed.
+ * @param start The position of the first item (of @p typesize size) from where data
+ * will be retrieved.
+ * @param nitems The number of items (of @p typesize size) that will be retrieved.
+ * @param dest The buffer where the decompressed data retrieved will be put.
```
FrancescAlted, line 454: Suggestion: "The buffer where the retrieved data will be stored (decompressed)."

blosc/blosc2.h
```diff
@@ -441,6 +441,24 @@ BLOSC_EXPORT int blosc2_chunk_repeatval(size_t nbytes, size_t typesize,
...
+ * @param destsize Size of the buffer where decompressed data received.
```
FrancescAlted, line 455: Suggestion: "Size of the buffer where retrieved data will be stored."

blosc/frame.c
FrancescAlted: +1

blosc/frame.c
FrancescAlted: +1

FrancescAlted (changes requested):
Looks good to me.  I did some suggestions on docstrings.  Feel free to use a better wording as you please.

And fully agree that we need better and more consistent endian handling in Blosc2.

blosc/blosc2.h
nmoinvaz: Done.

blosc/blosc2.h
nmoinvaz: Done.

blosc/blosc2.h
```diff
@@ -441,6 +441,24 @@ BLOSC_EXPORT int blosc2_chunk_repeatval(size_t nbytes, size_t typesize,
  */
 BLOSC_EXPORT int blosc_getitem(const void* src, int start, int nitems, void* dest);
 
+/**
+ * @brief Get @p nitems (of @p typesize size) in @p src buffer starting in @p start.
+ * The items are returned in @p dest buffer. The dest buffer should have enough space
+ * for storing all items. This function is a more secure version of #blosc_getitem.
+ *
+ * @param src The compressed buffer from data will be decompressed.
+ * @param srcsize Size of the compressed buffer from data will be decompressed.
```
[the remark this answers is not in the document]
nmoinvaz: Done.

blosc/blosc2.h
```diff
@@ -441,6 +441,24 @@ BLOSC_EXPORT int blosc2_chunk_repeatval(size_t nbytes, size_t typesize,
  */
 BLOSC_EXPORT int blosc_getitem(const void* src, int start, int nitems, void* dest);
 
+/**
+ * @brief Get @p nitems (of @p typesize size) in @p src buffer starting in @p start.
+ * The items are returned in @p dest buffer. The dest buffer should have enough space
+ * for storing all items. This function is a more secure version of #blosc_getitem.
+ *
+ * @param src The compressed buffer from data will be decompressed.
```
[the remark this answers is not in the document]
nmoinvaz: Done.

blosc/blosc2.c
nmoinvaz: It looks like there are a lot of places where reading `cbytes` or `nbytes` from a chunk, so I have used `blosc_cbuffer_sizes` instead. There are a few more places I need to get. Surprisingly this is almost all of the cases in the code where `+ BLOSC2_CHUNK_*` is used in `frame.c` or `schunk.c`  - for only getting those two values. There is a separate instance that checks blosc2_flags.

I would prefer not to expose `blosc_read_header` and `blosc_header_s` if necessary.

FrancescAlted (commented):
I like where you are headed, but I have two general comments.  Also, tests are not passing yet (but I suppose this is because the PR is a bit green yet).

blosc/frame.c
```diff
@@ -983,19 +1007,20 @@ uint8_t* get_coffsets(blosc2_frame_s *frame, int32_t header_len, int64_t cbytes,
     return NULL;
   }
 
-  int32_t coffsets_cbytes;
+  size_t coffsets_cbytes;
```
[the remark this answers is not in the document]
nmoinvaz: Fixed.

blosc/frame.c
```diff
@@ -1539,9 +1564,11 @@ blosc2_schunk* frame_to_schunk(blosc2_frame_s* frame, bool copy) {
   int64_t acc_nbytes = 0;
   int64_t acc_cbytes = 0;
   int32_t blocksize = 0;
-  int32_t csize = 0;
+  size_t chunk_nbytes;
+  size_t chunk_cbytes;
+  size_t chunk_blocksize;
```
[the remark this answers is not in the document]
nmoinvaz: Done. I have created `blosc2_cbuffer_sizes` with fixed `int32_t` return values and used it everywhere.

FrancescAlted (approved):
Awesome work!

PR #245, opened 2021-02-19
Fixed ASAN heap-buffer-overflow reading chunk outside of frame boundary.

nmoinvaz:
https://oss-fuzz.com/testcase-detail/6462229769748480


FrancescAlted:
LGTM

PR #250, opened 2021-02-24
Improve update_chunk for cframes

martaiborra:
Change `frame_update_chunk` so that when updating a chunk by a smaller one, the new one is placed at the same position and not at the end.

### DIFF
--- a/blosc/frame.c
+++ b/blosc/frame.c
@@ -2406,7 +2406,7 @@ void* frame_update_chunk(blosc2_frame_s* frame, int nchunk, void* chunk, blosc2_
       return NULL;
     }
 
-    if (chunk_old == 0) {
+    if (chunk_old == NULL) {
       cbytes_old = 0;
     }
     else {
--- a/blosc/frame.c
+++ b/blosc/frame.c
@@ -2402,7 +2402,8 @@ void* frame_update_chunk(blosc2_frame_s* frame, int nchunk, void* chunk, blosc2_
     uint8_t *chunk_old;
     int err = blosc2_schunk_get_chunk(schunk, nchunk, &chunk_old, &needs_free);
     if (err < 0) {
-      BLOSC_TRACE_ERROR("%d chunk con not obtenined from schunk.", nchunk);
+      BLOSC_TRACE_ERROR("%d chunk can not be obtained from schunk.", nchunk);
+      return NULL;
     }
 
     if (chunk_old == 0) {
--- a/blosc/frame.c
+++ b/blosc/frame.c
@@ -2393,6 +2393,31 @@ void* frame_update_chunk(blosc2_frame_s* frame, int nchunk, void* chunk, blosc2_
       return NULL;
     }
   }
+  int32_t cbytes_old;
+  int32_t old_offset;
+  if (!frame->sframe) {
+    // See how big would be the space
+    old_offset = offsets[nchunk];
+    bool needs_free;
+    uint8_t *chunk_old;
+    int err = blosc2_schunk_get_chunk(schunk, nchunk, &chunk_old, &needs_free);
+    if (err < 0) {
+      BLOSC_TRACE_ERROR("%d chunk con not obtenined from schunk.", nchunk);
+    }
+
+    if (chunk_old == 0) {
+      cbytes_old = 0;
+    }
+    else {
+      cbytes_old = sw32_(chunk_old + BLOSC2_CHUNK_CBYTES);
+      if (cbytes_old == BLOSC_MAX_OVERHEAD) {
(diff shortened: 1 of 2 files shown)


blosc/frame.c
```diff
@@ -2393,6 +2393,31 @@ void* frame_update_chunk(blosc2_frame_s* frame, int nchunk, void* chunk, blosc2_
...
+    // See how big would be the space
+    old_offset = offsets[nchunk];
+    bool needs_free;
+    uint8_t *chunk_old;
+    int err = blosc2_schunk_get_chunk(schunk, nchunk, &chunk_old, &needs_free);
+    if (err < 0) {
+      BLOSC_TRACE_ERROR("%d chunk con not obtenined from schunk.", nchunk);
+    }
+
+    if (chunk_old == 0) {
```
aleixalcacer, line 2408: It is better to use NULL with pointers.

blosc/frame.c
```diff
@@ -2393,6 +2393,31 @@ void* frame_update_chunk(blosc2_frame_s* frame, int nchunk, void* chunk, blosc2_
...
+      cbytes_old = 0;
+    }
+    else {
+      cbytes_old = sw32_(chunk_old + BLOSC2_CHUNK_CBYTES);
```
aleixalcacer, line 2412: Use `from_big` macro instead of `sw32_`.

blosc/schunk.c
```diff
@@ -686,6 +686,9 @@ int blosc2_schunk_update_chunk(blosc2_schunk *schunk, int nchunk, uint8_t *chunk
         if (frame->sframe) {
           schunk->cbytes -= cbytes_old;
         }
+        else if (cbytes_old >= cbytes) {
```
aleixalcacer, line 689: Maybe it's better to replace this with:
```
else {  // Contiguous frame
  if (cbytes_old >= cbytes) {
    schunk->cbytes -= cbytes;
  }
}
```

In this way, we visually divide the `sframe` and the `cframe` sections. Then, within the `cframe` section, we will already evaluate the condition.

What do you think?

*The `else if` statement is recommended to use in situations like:*

```
if (year < 18) {
  // Do something
} else if (18 >= year && year < 25) {
  // Do something
} else {
  // Do something
}
```

aleixalcacer (changes requested):

blosc/frame.c
```diff
@@ -2393,6 +2393,31 @@ void* frame_update_chunk(blosc2_frame_s* frame, int nchunk, void* chunk, blosc2_
...
+      if (cbytes_old == BLOSC_MAX_OVERHEAD) {
```
aleixalcacer, line 2413: Why?

blosc/frame.c
```diff
@@ -2421,6 +2446,10 @@ void* frame_update_chunk(blosc2_frame_s* frame, int nchunk, void* chunk, blosc2_
       }
   }
 
+  if (!frame->sframe && cbytes_chunk != 0 && cbytes_old >= cbytes_chunk) {
```
aleixalcacer, line 2449: Why `cbytes_chunk != 0`?

blosc/frame.c
```diff
@@ -2393,6 +2393,31 @@ void* frame_update_chunk(blosc2_frame_s* frame, int nchunk, void* chunk, blosc2_
...
+  int32_t cbytes_old;
+  int32_t old_offset;
+  if (!frame->sframe) {
+    // See how big would be the space
+    old_offset = offsets[nchunk];
+    bool needs_free;
+    uint8_t *chunk_old;
+    int err = blosc2_schunk_get_chunk(schunk, nchunk, &chunk_old, &needs_free);
+    if (err < 0) {
+      BLOSC_TRACE_ERROR("%d chunk con not obtenined from schunk.", nchunk);
```
aleixalcacer, line 2405: You must return an error code after tracing an error.

aleixalcacer:
Have you made sure that any test runs this?

blosc/frame.c
martaiborra: Thanks, fixed in 78ce44badf0b30b2fe0e3f4cb5451e39f26295cd

blosc/frame.c
martaiborra: Fixed in 9c3bf17a428af246f12b85fe2816bc0dd5177a98.
The `test_update_chunk` runs this in the case of a contiguous frame.

blosc/schunk.c
martaiborra: I see, changed in 9c3bf17a428af246f12b85fe2816bc0dd5177a98

aleixalcacer (approved):

PR #251, opened 2021-02-27
Fixed MSAN use-of-uninitialized when writing frame trailer

nmoinvaz:
Call to `fwrite` produces MSAN error because trailer is uninitialized.


FrancescAlted (approved):
Looks good.

PR #252, opened 2021-02-28
Fixed chunk not being freed when calling frame_special_chunk.

nmoinvaz:


FrancescAlted (approved):
Well spotted.  Thanks!

PR #255, opened 2021-03-03
Fixed MSAN error when doing pipeline filter if compressed stream token is zero.

nmoinvaz:
https://oss-fuzz.com/testcase-detail/6243118761508864
Added check for valid compressed stream token value.

### DIFF
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -1384,7 +1384,10 @@ static int blosc_d(
           return BLOSC2_ERROR_RUN_LENGTH;
         }
         uint8_t value = -cbytes;
-        memset(_dest, value, (unsigned int) neblock);
+        memset(_dest, value, (unsigned int)neblock);
+      } else {
+        BLOSC_TRACE_ERROR("Invalid or unsupported compressed stream token value - %d", token);
+        return BLOSC2_ERROR_RUN_LENGTH;
       }
       nbytes = neblock;
       cbytes = 0;  // everything is encoded in the cbytes token


nmoinvaz:
Interesting tid bit is that if `csize (aka cbytes) == 0` or `token == 0` then run is all zeros.

nmoinvaz:
Ping?

FrancescAlted:
Hmm, by looking at the code and https://github.com/Blosc/c-blosc2/blob/master/README_CHUNK_FORMAT.rst I think that `csize (aka cbytes) == 0` should have preference over `token == 0`; i.e. IIRC if `csize == 0`, there should be no token, and no `cdata` (this should be more explicit in the CHUNK_FORMAT doc).  If that happens, this should be considered a bug.  Could you open a ticket showing a minimal example so that it can be fixed?

nmoinvaz:
I don't see any in the code instance where there would be both `csize == 0` and `token == 0`. I was just noting that there are two ways to achieve the same thing.

FrancescAlted:
Ok. But the first (csize == 0) is shorter, so it is why it should have preference (which is they it was in the code, so not sure why you need fixing it that way; I'm probably missing something).

nmoinvaz:
Yes it might be shorter but according to the docs token == 0 is also valid so we need to account for that in the code. 

The fuzzer generated a file that hit that case. If you want to remove token == 0 from docs then that is also a valid option imo.

FrancescAlted:
https://github.com/Blosc/c-blosc2/pull/255/files#diff-3bd40d13b61e473400495c88e8db1b28b93426418383e6dcda98c7633ad1ed93L1361 should already acount for that. But agree that this should be better documented. Would that alone fix the issue?

FrancescAlted:
Improved docs in https://github.com/Blosc/c-blosc2/commit/2324c9516ee9ea321d312b28a644236aa9ee1d9d. Tell me if you find this still needs more editing (or if you find confident on how to improve it, go for it!).

nmoinvaz:
I think I may have been confused by `bit 0` thinking it was zero byte run, so it was my fault not understanding. Anyways the doc changes look good.  I have updated this PR.

FrancescAlted (approved):
Looks good!

PR #258, opened 2021-03-05
Add NDCELL filter

oscargm98:
The objective of this Pull Request is to add the new NDCELL filter


FrancescAlted:
This effort should be moved into the Caterva repository because it is meant for multdimensional datasets.  Once in Caterva, it must be using the new [API for user defined filters](https://github.com/Blosc/c-blosc2/pull/259).  I suggest using a branch in Caterva prior to make a PR.  @oscargm98 can you take care of this?

oscargm98:
> This effort should be moved into the Caterva repository because it is meant for multdimensional datasets. Once in Caterva, it must be using the new [API for user defined filters](https://github.com/Blosc/c-blosc2/pull/259). I suggest using a branch in Caterva prior to make a PR. @oscargm98 can you take care of this?

Of course, it will be moved to Caterva

PR #264, opened 2021-03-13
Fixed ASAN null-deference read trying to free chunk 

nmoinvaz:
https://oss-fuzz.com/testcase-detail/5107303620673536


FrancescAlted (approved):

nmoinvaz:
Also fixes:

https://oss-fuzz.com/testcase?key=5994237083648000
https://oss-fuzz.com/testcase?key=5099327329533952
https://oss-fuzz.com/testcase?key=5788172974948352

PR #266, opened 2021-03-13
Fixed ASAN heap-buffer-overflow when reading frame chunk cbytes

nmoinvaz:
https://oss-fuzz.com/testcase-detail/5206670272757760

### DIFF
--- a/blosc/frame.c
+++ b/blosc/frame.c
@@ -2018,12 +2018,20 @@ int frame_get_lazychunk(blosc2_frame_s *frame, int nchunk, uint8_t **chunk, bool
     free(block_csizes);
   } else {
     // The chunk is in memory and just one pointer away
-    *chunk = frame->cframe + header_len + offset;
-    if ((int64_t)header_len + offset + BLOSC_MIN_HEADER_LENGTH > frame->len) {
+    int64_t chunk_header_offset = header_len + offset;
+    int64_t chunk_cbytes_offset = chunk_header_offset + BLOSC_MIN_HEADER_LENGTH;
+
+    *chunk = frame->cframe + chunk_header_offset;
+
+    if (chunk_cbytes_offset > frame->len) {
       BLOSC_TRACE_ERROR("Cannot read the header for chunk in the (contiguous) frame.");
       rc = BLOSC2_ERROR_READ_BUFFER;
     } else {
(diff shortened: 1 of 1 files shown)


blosc/frame.c
```diff
@@ -2018,12 +2018,20 @@ int frame_get_lazychunk(blosc2_frame_s *frame, int nchunk, uint8_t **chunk, bool
...
+
+    if (chunk_cbytes_offset > frame->len) {
       BLOSC_TRACE_ERROR("Cannot read the header for chunk in the (contiguous) frame.");
       rc = BLOSC2_ERROR_READ_BUFFER;
     } else {
       rc = blosc2_cbuffer_sizes(*chunk, NULL, &lazychunk_cbytes, NULL);
+      if (rc && chunk_cbytes_offset + lazychunk_cbytes > frame_len) {
+        BLOSC_TRACE_ERROR("Compressed bytes exceed beyond frame length.");
+        rc = BLOSC2_ERROR_READ_BUFFER;
+      }
```
FrancescAlted, line 2034: Is this necessary for fixing the issue? IIRC lazy chunks are built dynamically during retrievals and are not part of the frame, so I'd say this check is not correct.

nmoinvaz: This is part of the retrieval process. Reading the chunk and then verifying that the cbytes don't exceed frame length? Or is there another value I should be checking against rather than frame length?

FrancescAlted: What I mean is that `lazychunk_cbytes` is the length of the lazy chunk, not the actual chunk. Frames are made of actual chunks. So unless I'm missing something I think this check should use the length of the *actual* chunk.

nmoinvaz: I think perhaps the name of the variable is wrong? Because it says in the comment above:

`The chunk is in memory and just one pointer away`

So maybe it should just be called `chunk_cbytes`. Lazy chunks I assume can only be on disk and are handled by the `if (frame->cframe == NULL)` branch. But my understanding may be off.

FrancescAlted: Ah, I see where the confusion gets from.  So when a chunk has to be retrieved from memory there is no point to build a lazy chunk out of it, so we are fetching the chunk and updating the `lazychunk_cbytes` (because it needs to be returned).  Sorry that I got confused because of the naming (sometimes is difficult to review something in small phone screens without easy access to context).  I think the PR is correct then.

PR #269, opened 2021-03-16
Btune plugin

aleixalcacer:
This implements a plugin capability for users to specify their own tuning functions for adjusting compression parameters like `clevel`, `compressor`, `shuffle/bitshuffle` and others.  This allows for dynamically doing that while e.g. doing appends into a super-chunk, so that the compression parameters can change (tuned) from one chunk to another.

### DIFF
--- a/blosc/CMakeLists.txt
+++ b/blosc/CMakeLists.txt
@@ -50,7 +50,7 @@ endif()
 include_directories(${BLOSC_INCLUDE_DIRS})
 
 # library sources
-set(SOURCES blosc2.c blosc2-common.h blosclz.c fastcopy.c fastcopy.h schunk.c frame.c btune.c btune.h
+set(SOURCES blosc2.c blosc2-common.h blosclz.c fastcopy.c fastcopy.h schunk.c frame.c stune.c stune.h
         context.h delta.c delta.h shuffle-generic.c bitshuffle-generic.c trunc-prec.c trunc-prec.h
         timestamp.c sframe.c directories.c)
 if(NOT CMAKE_SYSTEM_PROCESSOR STREQUAL arm64)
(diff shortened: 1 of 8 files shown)


FrancescAlted (approved):
Looks good to me.

FrancescAlted:
This looks great, and should ready for merging.

However, we have decided here to change the way how the pipeline metainfo is serialized in frames.  In brief, the position (but not the order) of the different filters was altered during the creation of a frame, so after re-opening the user could see that the positions were not the same.  For the sake of the principle of minimal surprise we decided to fix this here.  The only issue that we need to rebuild the file frames with the 'old' format in https://github.com/Blosc/c-blosc2/tree/master/tests/fuzz/corpus.  @nmoinvaz can you shed some light on how to do this?  Also, it would be a good idea to document how to do that (a README in the `corpus` directory?). Thanks!

FrancescAlted:
We decided to remove the condition for having the shuffle/bitshuffle filter at the end of the filter pipeline that we added in this PR.  So this would not break backward compatibility and can be merged as-is.

On the other hand, I still think that documenting the way on how the files in the `fuzz/corpus/` are generated would be good.  Actually it would be nice if you can put the generator for those in the `fuzz/` directory, together with some README file with instructions.

nmoinvaz:
@FrancescAlted I'm not sure why I didn't see this message earlier. To generate those I believe I used `test_simple` and just dumped the frames to a file after they were written in memory. I don't have the code I used to generate them although it would be good to have a script or program that did generate them.

FrancescAlted:
> @FrancescAlted I'm not sure why I didn't see this message earlier. To generate those I believe I used `test_simple` and just dumped the frames to a file after they were written in memory. I don't have the code I used to generate them although it would be good to have a script or program that did generate them. When I get some time I can look into it.

Done in https://github.com/Blosc/c-blosc2/commit/fff2146e7e30e7670b755e3f72dc2809bb61b848.

nmoinvaz:
It looks really good!

PR #273, opened 2021-04-06
Input/output plugin

aleixalcacer:


blosc/blosc2.h
```diff
@@ -711,6 +731,24 @@ BLOSC_EXPORT void blosc_cbuffer_versions(const void* cbuffer, int* version,
  */
 BLOSC_EXPORT const char* blosc_cbuffer_complib(const void* cbuffer);
 
+/*********************************************************************
+  Structures and functions related with user-defined input/output.
+*********************************************************************/
+typedef struct {
+  void* (*open)(const char *urlpath, const char *mode, void *params);
```
nmoinvaz, line 738: Consider using typedefs for these callbacks.
```
typedef void* (*blosc2_open_cb)(const char *urlpath, const char *mode, void *params);
typedef int (*blosc2_close_cb)(void *stream, void* params);
```
Then in your struct you just have:
```
typedef struct {
  blosc2_open_cb open;
  blosc2_close_cb close;
  ...
} blosc2_io;
```

nmoinvaz: If you look in my minizip-ng project I have done something similar:
https://github.com/zlib-ng/minizip-ng/blob/master/mz_strm.h

blosc/frame.h
```diff
@@ -56,6 +56,7 @@ typedef struct {
   int64_t maxlen;           //!< The maximum length of the frame; if 0, there is no maximum
   uint32_t trailer_len;     //!< The current length of the trailer in (compressed) bytes
   bool sframe;              //!< Whether the frame is sparse (true) or not
+  blosc2_schunk *schunk;     //!< The schunk associated
```
nmoinvaz, line 59: Comment is not aligned?

blosc/schunk.c
```diff
@@ -354,9 +356,9 @@ int64_t blosc2_schunk_to_buffer(blosc2_schunk* schunk, uint8_t** dest, bool* nee
 
 /* Write an in-memory frame out to a file. */
 int64_t frame_to_file(blosc2_frame_s* frame, const char* urlpath) {
-  FILE* fp = fopen(urlpath, "wb");
-  size_t nitems = fwrite(frame->cframe, (size_t)frame->len, 1, fp);
-  fclose(fp);
+  FILE* fp = frame->schunk->storage->udio->open(urlpath, "wb", NULL);
```
nmoinvaz, line 359: Will open always return `FILE *`? Perhaps it should be `void *fp`. Because if you have a memory stream, it might not return `FILE *` but would instead return a pointer to internal struct about memory allocation etc.

examples/compress_file.c
```diff
@@ -30,6 +30,7 @@
 
 
 int main(int argc, char* argv[]) {
+  blosc_init();
```
nmoinvaz, line 33: I'm curious about why the calls to `blosc_init` and `blosc_destroy`. I thought it is not necessary to do those calls. Perhaps my knowledge of it is old?

aleixalcacer: I used the `blosc_init` to initialize a global variable for the io plugin, but it is removed in the last commit.

Although `blosc_init` and `blosc_destroy` are not necessary, I think that their usage should be recommendable (`blosc_init` initializes some structures):

https://github.com/Blosc/c-blosc2/blob/f489cdbd0a2567e07365671618fd859a31eec4ae/blosc/blosc2.c#L3241-L3252

@FrancescAlted ?

blosc/schunk.c
aleixalcacer: Nop, you are right. This plugin is specially designed to manage different types of io formats (s3, db ...). Thanks!

blosc/frame.h
aleixalcacer: Fixed

blosc/blosc2.h
aleixalcacer: I agree, thank you very much for the suggestion!

aleixalcacer:
@nmoinvaz can you take a quick look at the 2 checks that are failing? I would appreciate a hint on what's going on :)

blosc/blosc2.h
```diff
@@ -20,9 +20,29 @@
...
+  #include <malloc.h>
+
+/* stdint.h only available in VS2010 (VC++ 16.0) and newer */
+  #if defined(_MSC_VER) && _MSC_VER < 1600
+    #include "win32/stdint-windows.h"
+  #else
+    #include <stdint.h>
+  #endif
+
+  #define fseek _fseeki64
```
nmoinvaz, line 38: Based on the definition of [_fseeki64](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fseek-fseeki64?view=msvc-160) using `__int64` your definition of `blosc2_seek_cb` might have a problem.

examples/compress_file.c
FrancescAlted: Yes.  `blosc_init()` and `blosc_destroy()` were introduced initially in C-Blosc (1) so as to be able to initialize global structures.  C-Blosc2 uses contexts, so the usefulness of these is more limited now.  But we want to keep API compatibility with C-Blosc (1), so we need to keep them around in C-Blosc2 too.

blosc/blosc2.h
aleixalcacer: In this case, It seems this is the problem :)

Otherwise, this definition is also used in the `master` branch:

https://github.com/Blosc/c-blosc2/blob/f489cdbd0a2567e07365671618fd859a31eec4ae/blosc/blosc2.c#L71

Maybe we have found why are some random failures on Win32.

@FrancescAlted, do you know the reasons for introducing this definition?

FrancescAlted: IIRC I have introduced `_fseeki64` so as to support files larger than 2 GB on Win32.  But obviously this has some undesirable consequences (see also https://stackoverflow.com/questions/4003405/32-bit-windows-and-the-2gb-file-size-limit-c-with-fseek-and-ftell).  In addition, Microsoft is not providing Win32 for OEMS for a year now (https://www.engadget.com/windows-10-32-bit-oem-173055990.html).  Hence, if there is no opposition, I very much prefer to drop the support of Win32 entirely.

nmoinvaz: @FrancescAlted I'm not sure if `_fseeki64` has anything to do with Win32 or Win64. I always believed that `fseek` on Windows only allowed 2GB offset and `_fseeki64` allowed greater than 2GB offset regardless of the subsystem being used. The size of long (for offset arg) also does not change when compiling on 32-bit Windows or 64-bit Windows afaik. The original code may be correct, but we may need to find another way to do it.. like change the definition of `blosc2_seek_cb` when `fseek64i` is used.

FrancescAlted: Ok, thanks. Maybe @aleixalcacer would want to have another attempt at using int64_t, but if that doesn't work, I would go ahead and drop win32 support.

blosc/blosc2.h
```diff
@@ -711,6 +734,31 @@ BLOSC_EXPORT void blosc_cbuffer_versions(const void* cbuffer, int* version,
...
+typedef struct {
+  blosc2_open_cb open;
+  blosc2_close_cb close;
+  blosc2_seek_cb seek;
+  blosc2_write_cb write;
+  blosc2_read_cb read;
+} blosc2_io;
+
+static const blosc2_io BLOSC2_IO_DEFAULTS = {
+  .open = (blosc2_open_cb) fopen,
```
nmoinvaz, line 756: Because of the integer type problem in `fseek` for `offset` it would be better to have a custom wrapper of fseek that is called. For example, on 32-bit Windows when `fseek` is called it will still use `long int` which I believe is 4 bytes and so the function signature will not match with `blosc2_seek_cb`.

It might be best to do this for all the FILE functions. Even if they only wrap a single call. That way you can separate your interface from the interface of the FILE functions which vary on the different platforms. 

https://github.com/zlib-ng/minizip-ng/blob/master/mz_strm_os_posix.c

aleixalcacer: I have implemented it in f5846e5. Is it what you suggested?

FrancescAlted (approved):
Nice work!  Looks really good to me.

FrancescAlted:
I am all about simplifying things :-). Many of the code in c-blosc2 inherits from the original c-blosc(1) sources that date from ~10 years ago. @nmoinvaz I'd encourage you to propose a PR with the minimum MSVC version.  Thanks in advance!

nmoinvaz:
@aleixalcacer there are some `#define fseek _fseeki64 ` at the top of some of the source files. I am thinking perhaps those are no longer needed now.

aleixalcacer:
The `params` parameter has been removed from any function (except `open`) as suggested by @nmoinvaz.

The reason for not removing this parameter from "open" is because they may be required within the `open` in some third-party use cases. (For example, authentication parameters would be required to "open" a remote file).

I can't find another way to do it 😅

blosc/blosc2.h
```diff
@@ -736,26 +726,32 @@ BLOSC_EXPORT const char* blosc_cbuffer_complib(const void* cbuffer);
...
 
 typedef struct {
   blosc2_open_cb open;
   blosc2_close_cb close;
+  blosc2_tell_cb tell;
   blosc2_seek_cb seek;
   blosc2_write_cb write;
   blosc2_read_cb read;
+  blosc2_truncate_cb truncate;
   void *params;
```
nmoinvaz, line 744: Looks you removed `params` from all the arguments which is good. Also probably need to remove `params` in i/o interface the struct too. Having `params` in open callback is ok.

aleixalcacer: The `params` in the `blosc_io` structure are needed to allow the users to set their params. See an example of use: https://github.com/Blosc/c-blosc2/blob/86eaeba547eb5e15536e8dda7135957191fb1e3b/tests/test_udio.c

Another option may be to split the `blosc_io` structure in two: `blosc_io` (only callbacks) and `blosc_io_params`?

nmoinvaz: Sorry for the delay. It is a unusual design pattern for the `io_params` in `test_udio.c`. I was thinking you can return `io_params` with an accessor method like:
```
struct test_udio_params* test_get_udio_params(void *stream) {
   test_file *my = (test_file *) stream;
   return my->params;
}
```
Then you'd expose the stream handle from `blosc2_schunk` or something like that. But since the stream can get created and deleted multiple times it doesn't make much sense.

I think you are probably better off creating a global `test_udio_params` variable in your `test_udio.c` and removing `void *params` from the `open` signature and from the interface.

FrancescAlted: We have changed a bit the way to register the plugins (so we don't need to pass the callbacks to the storage).  We are going to merge this as-is.  In case you still have a better suggestion, please feel free to send it as a PR.

FrancescAlted:
Looks good to me!

nmoinvaz:
Nice work. I saw that you even made some changes to the stdint.h check so that is good. I was going to do it but looks like you already did it another way so that works too.

PR #274, opened 2021-04-09
Postfilters runlen

FrancescAlted:
This fixes the execution of postfilters even when special chunks (zeros, constants, NaNs) are present.

### DIFF
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -1225,7 +1225,9 @@ int pipeline_d(struct thread_context* thread_context, const int32_t bsize, uint8
 
 
 int32_t set_nans(int32_t typesize, uint8_t* dest, int32_t destsize) {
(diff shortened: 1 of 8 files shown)


blosc/blosc2.c
```diff
@@ -1223,17 +1224,90 @@ int pipeline_d(struct thread_context* thread_context, const int32_t bsize, uint8
...
+  return BLOSC2_ERROR_DATA;
+}
+
+
+int32_t set_values(int32_t typesize, const uint8_t* src, uint8_t* dest, int32_t destsize) {
+  int32_t nitems = destsize / typesize;
+  if (nitems > destsize / typesize) {
+    nitems = destsize / typesize;
+  }
+
```
nmoinvaz, line 1260: What happens when `nitems == 0`?

FrancescAlted: Fixed.

PR #275, opened 2021-04-11
Added max blocksize with check for 32-bit context allocation

nmoinvaz:
> Added maximum block size to prevent context tmp buffer from overflowing in 4 * ebsize calculation. Maximum size is 512mb - 4k which is closest in value to (INT32_MAX / 4) - (BLOSC_MAX_TYPESIZE * sizeof(int32_t)). This allows 32-bit apps to allocate enough memory without tmp_nbytes overflowing during context creation.

Should solve `clusterfuzz-testcase-decompress_frame_fuzzer-4516447050137600` and probably some others.

### DIFF
--- a/blosc/blosc2.h
+++ b/blosc/blosc2.h
@@ -110,7 +110,7 @@ enum {
   BLOSC_BITSHUFFLE = 2,  //!< bit-wise shuffle
   BLOSC_DELTA = 3,       //!< delta filter
   BLOSC_TRUNC_PREC = 4,  //!< truncate precision filter
-  BLOSC_LAST_FILTER= 5,  //!< sentinel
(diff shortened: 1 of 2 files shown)


blosc/blosc2.h
```diff
@@ -141,6 +141,7 @@ enum {
  */
 enum {
   BLOSC2_MAXDICTSIZE = 128 * 1024, //!< maximum size for compression dicts
+  BLOSC2_MAXBLOCKSIZE = 536866816  //!< maximum size for blocks
```
FrancescAlted, line 144: I prefer to see something more readable like, "512 * 1024 * 1024", but not too important ;-)

FrancescAlted (approved):
LGTM

PR #276, opened 2021-04-12
Add delete_chunk function

aleixalcacer:


tests/test_delete_chunk.c
```diff
@@ -0,0 +1,178 @@
...
+typedef struct {
+  bool contiguous;
+  char *urlpath;
+}test_storage;
+
+test_storage tstorage[] = {
+    // {false, NULL},  // memory - schunk
+    // {true, NULL},  // memory - cframe
+    // {true, "test_delete_chunk.b2frame"}, // disk - cframe
+    {false, "test_delete_chunk_s.b2frame"}, // disk - sframe
```
FrancescAlted, line 54: You probably want to uncomment the three cases above.

tests/test_delete_chunk.c
```diff
@@ -0,0 +1,178 @@
+/*
+  Copyright (C) 2020- The Blosc Development Team <blosc@blosc.org>
+  http://blosc.org
+  License: BSD (see LICENSE.txt)
+
+  Creation date: 2020-09-23
```
FrancescAlted, line 6: Not the correct creation date.  Hmmm, probably worth getting rid of this line for future files.

FrancescAlted (changes requested):
Needs to check for more cases (see comments).

blosc/frame.c
```diff
@@ -2645,6 +2645,142 @@ void* frame_update_chunk(blosc2_frame_s* frame, int nchunk, void* chunk, blosc2_
...
+  for (int i = nchunk; i < nchunks - 1; i++) {
+    offsets[i] = offsets[i + 1];
+  }
+  offsets[nchunks - 1] = 0;
+
+  // Re-compress the offsets again
+  blosc2_context* cctx = blosc2_create_cctx(BLOSC2_CPARAMS_DEFAULTS);
+  cctx->typesize = 8;
+  void* off_chunk = malloc((size_t)off_nbytes + BLOSC_MAX_OVERHEAD);
+  int32_t new_off_cbytes = blosc2_compress_ctx(cctx, offsets, off_nbytes - 8,
```
FrancescAlted, line 2700: For clarity, I suggest to use `- sizeof(int64_t)` instead of `- 8`.

FrancescAlted (approved):

PR #277, opened 2021-04-12
Let decompress an empty array

martaiborra:
For python-blosc2 we may want to compress an empty array.
The problem is that `blosc_decompress` returns zero or a negative value if an errors occurs, otherwise it returns the number of bytes decompressed. Which in the case of an empty array would be 0.

If in the python-blosc2 ` decompress `function the 0 return value is ignored when decompressing, empty arrays could be compressed and decompressed... The other solution is to change the function from `blosc2` to only return a negative value when an error occurs.

### DIFF
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -692,7 +692,7 @@ int read_chunk_header(const uint8_t* src, int32_t srcsize, bool extended_header,
     BLOSC_TRACE_ERROR("`cbytes` is too small to read min header.");
     return BLOSC2_ERROR_INVALID_HEADER;
   }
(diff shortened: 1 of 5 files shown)


FrancescAlted:
That's a good point.  For consistency, I feel that we should support compression (and decompression) of buffers of 0 bytes.  I'd add a test (you can use this branch for that) for a zero-length buffer and try to make it work.  If you succeed, we should update the docstrings (blosc2.h) so that returning a zero when compressing/decompressing is not an error anymore.  Could you please try to explore this way?

FrancescAlted (approved):
Looks good to me!

PR #278, opened 2021-04-12
Special chunks

FrancescAlted:
This introduces a new special chunk for uninitizalized values.  This does not take time to decompress because the data is not copied anyware, and it is meant for quickly creating large super-chunks with uninitialized values.

As bonus, this PR reserves space for 3 additional special values to be used in the future.

### DIFF
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -3495,31 +3495,27 @@ int blosc2_set_maskout(blosc2_context *ctx, bool *maskout, int nblocks) {
 
 
 /* Create a chunk made of zeros */
-int blosc2_chunk_zeros(const size_t nbytes, const size_t typesize, void* dest, size_t destsize) {
+int blosc2_chunk_zeros(blosc2_cparams cparams, const size_t nbytes, void* dest, size_t destsize) {
   blosc_header header;
+  blosc2_context* context = blosc2_create_cctx(cparams);
 
-  if (typesize == 0 || typesize > BLOSC_MAX_TYPESIZE) {
-    BLOSC_TRACE_ERROR("typesize cannot be larger than %d bytes", BLOSC_MAX_TYPESIZE);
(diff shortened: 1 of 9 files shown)


blosc/blosc2.c
```diff
@@ -1963,12 +1966,28 @@ static int initialize_context_decompression(blosc2_context* context, blosc_heade
...
 
-  context->runlen_type = (header->blosc2_flags >> 4) & BLOSC2_RUNLEN_MASK;
+  context->special_type = (header->blosc2_flags >> 4) & BLOSC2_SPECIAL_MASK;
+  if (context->special_type > BLOSC2_SPECIAL_LASTID) {
+    BLOSC_TRACE_ERROR("Unknown special values ID (%d) ",
+                      context->special_type);
+    return BLOSC2_ERROR_DATA;
+  }
+
+  int memcpy = (context->header_flags & (uint8_t) BLOSC_MEMCPYED);
```
aleixalcacer, line 1976: Should not use function names as variable names.

blosc/frame.c
```diff
@@ -995,6 +998,12 @@ uint8_t* get_coffsets(blosc2_frame_s *frame, int32_t header_len, int64_t cbytes,
         BLOSC_TRACE_ERROR("Cannot read the cbytes outside of frame boundary.");
         return NULL;
       }
+      if (chunk_nbytes != nchunks * sizeof(int64_t)) {
```
aleixalcacer, line 1001: Is it necessary? (it changes the function signature)

blosc/blosc2.c
FrancescAlted: Fixed in 3dedb12

blosc/frame.c
FrancescAlted: This is a good consistency check, that helps to find malformed meta-info.  So it should stay.

PR #298, opened 2021-05-08
Fixed ASAN heap-buffer-overflow when reading frame metalayers.

nmoinvaz:
Similar to #279. Variable `header_pos` is used for checking boundaries when reading name/offsets. Don't need to increase `header_pos` when seeking to another offset in the header to read metadata.

https://oss-fuzz.com/testcase-detail/5361858753200128


FrancescAlted (approved):
LGTM

PR #299, opened 2021-05-10
Added typesize check when reading frame header info.

nmoinvaz:
https://oss-fuzz.com/testcase-detail/6103287616110592

Also I have moved some length checks closer to where the values are read. This should prevent it from reading fields if header length is too small.

### DIFF
--- a/blosc/schunk.c
+++ b/blosc/schunk.c
@@ -469,8 +469,8 @@ blosc2_schunk* blosc2_schunk_from_buffer(uint8_t *cframe, int64_t len, bool copy
     return NULL;
   }
   blosc2_schunk* schunk = frame_to_schunk(frame, copy, &BLOSC2_IO_DEFAULTS);
-  if (copy) {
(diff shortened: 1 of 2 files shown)


nmoinvaz:
I forgot I had this open. Hopefully I can look at it this next week.

blosc/schunk.c
```diff
@@ -469,8 +469,8 @@ blosc2_schunk* blosc2_schunk_from_buffer(uint8_t *cframe, int64_t len, bool copy
     return NULL;
   }
   blosc2_schunk* schunk = frame_to_schunk(frame, copy, &BLOSC2_IO_DEFAULTS);
-  if (copy) {
-    // We don't need the frame anymore
+  if (schunk && copy) {
+    // Super-chunk has its own copy of frame
```
FrancescAlted, line 473: Well spotted.

FrancescAlted (approved):
LGTM

PR #303, opened 2021-05-14
Fix mkdir bug for MinGW-w64 gcc

martaiborra:
With the MinGW-w64 gcc compiler there is an incompatibility due to the mkdir function.

Probably, it will be needed to add `-DCMAKE_RC_COMPILER=windres` as CMake Options in the settings to build the project.

### DIFF
--- a/blosc/schunk.c
+++ b/blosc/schunk.c
@@ -18,9 +18,12 @@
 #include "frame.h"
 #include "stune.h"
 
-#if defined(_WIN32) && !defined(__MINGW32__)
+#if defined(_WIN32)
   #include <windows.h>
+  #include <direct.h>
   #include <malloc.h>
+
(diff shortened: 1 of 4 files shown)


aleixalcacer:
@nmoinvaz, we have had some problems compiling `zlib-ng` with MinGW on Windows. Finally, we have solved it using this cmake flag: `-DCMAKE_RC_COMPILER=windres`. Do you know if it is the correct way to do it?

blosc/schunk.c
```diff
@@ -34,6 +31,12 @@
 
 #endif  /* _WIN32 */
 
+#if defined(_WIN32)
```
FrancescAlted, line 34: This includes MSVC, which currently works.  I think a better solution would be to add another condition so that the `mkdir` redefinition only happens on MinGW.

Issue #306, opened 2021-05-17
I need help with supported platforms


waralex:
This is https://github.com/JuliaPackaging/Yggdrasil/pull/3001 PR with script for building shared lib of blosc2 v2.0.0rc1 for using in Julia packages. 
It tries to build blosc2 on all platforms that supported by Julia
The build does not pass on Windows for an obvious reason (https://github.com/Blosc/c-blosc2/issues/302)
Also, the build does not pass on all `arm` architectures, and here I need your help. Is blosc2 in the principle not compilable in arm, or do I just need to change the build script for it?
Right now I use the same build script for all platforms:
```
cd $WORKSPACE/srcdir
cd c-blosc2/
mkdir build
cd build/
cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTS=OFF -DBUILD_BENCHMARKS=OFF -DBUILD_EXAMPLES=OFF -DBUILD_STATIC=OFF ..
make -j${nproc}
make install
install_license ../LICENSES/*.txt
```

Errors during building on arm platforms can be viewed in the `checks` in PR.

Thank you in advance for your help!

FrancescAlted:
Hi @waralex.  I have been tweaking the cmake config file so that ARM platform could be better supported.  Can you give it another go? Regarding #302, we are still trying to figure out a solution (we are having a problem with our local MSVC 2019 instance). Thanks for the effort!

FrancescAlted:
Hi.  I have double checked that our cmake config works just fine on arm64 on both Mac Silicon and on an `arm64v8` running Ubuntu latest.  Here it is an example of cmake output with Ubuntu:

```
$ mkdir build
$ cd build
$ cmake ..
-- The C compiler identification is GNU 9.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
Configuring for Blosc version: 2.0.0.rc.2.dev
-- Using LZ4 internal sources.
-- Using ZLIB-NG internal sources for ZLIB support.
-- Using CMake version 3.18.4
-- ZLIB_HEADER_VERSION: 1.2.11
-- ZLIBNG_HEADER_VERSION: 2.0.3
[...]

PR #307, opened 2021-05-19
Patch armv7l shuffle/bitshuffle support include header, lax-vector-conv

mkitti:
This patch resolves the armv7l build issues identified in #306

### DIFF
--- a/blosc/CMakeLists.txt
+++ b/blosc/CMakeLists.txt
@@ -194,7 +194,7 @@ if(COMPILER_SUPPORT_NEON)
         # Only armv7l needs special -mfpu=neon flag; aarch64 doesn't.
       set_source_files_properties(
             shuffle-neon.c bitshuffle-neon.c
-            PROPERTIES COMPILE_FLAGS "-mfpu=neon")
+            PROPERTIES COMPILE_FLAGS "-mfpu=neon -flax-vector-conversions")
     endif()
     # Define a symbol for the shuffle-dispatch implementation
     # so it knows NEON is supported even though that file is
--- a/blosc/shuffle.c
+++ b/blosc/shuffle.c
@@ -36,6 +36,7 @@
 #endif  /* defined(SHUFFLE_SSE2_ENABLED) */
 
 #if defined(SHUFFLE_NEON_ENABLED)
+  #include <sys/auxv.h>
   #include "shuffle-neon.h"
   #include "bitshuffle-neon.h"
 #endif  /* defined(SHUFFLE_NEON_ENABLED) */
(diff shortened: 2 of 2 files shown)


mkitti:
1. Subsequent `set_source_files_properties` calls in CMake override previous calls, so `-flax-vector-conversions` needs to be repeated.
2. `#include <sys/auxv.h>` is needed for `getauxval(AT_HWCAP) & HWCAP_ARM_NEO`.

The above allows crosscompiling via BinaryBuilder.

mkitti:
Update coming on this...

mkitti:
Updated

mkitti:
Our BinaryBuilder crosscompiler builds on all of our supported platforms: https://github.com/JuliaPackaging/Yggdrasil/runs/2624565172

mkitti:
I added 54654ac since the prior commit was failing on aarch64-apple-darwin, which is still experimental for us.

https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=11102&view=logs&jobId=b73de0a3-cc76-51c8-66e8-3220d54371bd&j=b73de0a3-cc76-51c8-66e8-3220d54371bd&t=f2334286-2aa2-54d4-6f03-edd7fb9da34e

mkitti:
The code in https://github.com/Blosc/c-blosc2/blob/c7071132b69d976510c255b8c1a823a3516f0d54/internal-complibs/zlib-ng-2.0.3/arch/arm/armfeature.c looks a bit more robust in terms of detecting arm features.

In particular, the Linux specific code (`sys/auxv.h`, `getauxval(AT_HWCAP) & HWCAP_ARM_NEON` ) should be included only if `defined(__linux__)`.

mkitti:
You might also need some version of `arm_has_neon` there:

This block looks like it supports older Linux:
https://github.com/Blosc/c-blosc2/blob/c7071132b69d976510c255b8c1a823a3516f0d54/internal-complibs/zlib-ng-2.0.3/arch/arm/armfeature.c#L45-L49

This block supports non-aarch64 Apple on ARM (aarch32?):
https://github.com/Blosc/c-blosc2/blob/c7071132b69d976510c255b8c1a823a3516f0d54/internal-complibs/zlib-ng-2.0.3/arch/arm/armfeature.c#L51-L54

FrancescAlted:
I am not sure we should touch this because right now the test suite is passing in all your ARM platforms, right?  The only exception is aarch64 on Apple, but then this is interesting because I cannot reproduce that in my MacBook M1 box, which, if I am not wrong looks very similar to yours:

```
> uname -a
Darwin Francescs-MacBook-Air.local 20.4.0 Darwin Kernel Version 20.4.0: Fri Mar  5 01:14:02 PST 2021; root:xnu-7195.101.1~3/RELEASE_ARM64_T8101 arm64
```

But in case you find they are useful, for all sake send a PR.

mkitti:
Yes, currently everything we are building is passing now that you removed the header, so I do not there is nothing very urgent.

I suspect the issue with the header was using the non-Apple provided `Clang` (as opposed to `AppleClang`) and setting `_XOPEN_SOURCE=600` due to the code below.

https://github.com/Blosc/c-blosc2/blob/a723ef86984309b93a14dc1088fba0b2e61177b4/CMakeLists.txt#L315-L318

I personally do not have any Apple hardware to test on, so I'm reticent to purpose major changes in that direction due the lack of my ability to actually test it beyond looking at the CI.

PR #312, opened 2021-06-02
Added info about fuzz testing for developers.

nmoinvaz:


FrancescAlted:
This is great.  Do you plan to do more work on this or we can merge it already?

nmoinvaz:
Ir is ready unless you can think of anything more I should mention.

FrancescAlted:
I find this really cool as it is now, so merging.

PR #322, opened 2021-06-20
Added plugin input check

nmoinvaz:
Also fixed `HAVE_PLUGINS` not being defined.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -128,12 +128,12 @@ if(MINGW)
     endif()
 endif(MINGW)
 
-if (ENABLE_ASAN)
+if(ENABLE_ASAN)
     message(STATUS "Enabling ASAN")
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g -Os -fno-omit-frame-pointer -fsanitize=address")
 endif()
 
-if (BUILD_PLUGINS AND BUILD_LITE)
+if(BUILD_PLUGINS AND BUILD_LITE)
     message(FATAL_ERROR "Both plugins and lite Blosc versions can not be built at the same time.")
 endif()
 
@@ -166,7 +166,7 @@ if(NOT DEACTIVATE_ZLIB)
     set(HAVE_ZLIB TRUE)
 endif()
 
-if (HAVE_ZLIB_NG)
+if(HAVE_ZLIB_NG)
     set(ZLIB_NG_DIR "zlib-ng-2.0.3")  # update to the actual included version
     add_definitions(-DZLIB_COMPAT)
     set(ZLIB_COMPAT TRUE)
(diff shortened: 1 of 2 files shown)


nmoinvaz:
Odd that that one test is failing only on Win64..

CMakeLists.txt
```diff
@@ -132,16 +132,16 @@ if(MINGW)
     endif()
 endif(MINGW)
 
-if (ENABLE_ASAN)
+if(ENABLE_ASAN)
     message(STATUS "Enabling ASAN")
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g -Os -fno-omit-frame-pointer -fsanitize=address")
 endif()
 
-if ((BUILD_PLUGINS STREQUAL ON) AND (BUILD_LITE STREQUAL ON))
```
oscargm98, line 140: I changed this and made use of STREQUAL because, in my local c-blosc2, when I do not use STREQUAL the condition does activate if BUILD_PLUGINS and BUILD_LITE are both OFF, but we want the condition to be TRUE only if  BUILD_PLUGINS and BUILD_LITE are both ON

CMakeLists.txt
```diff
@@ -215,7 +215,7 @@ if(NOT DEACTIVATE_IPP)
 endif()
 
 if(BUILD_PLUGINS)
-    set(HAVE_PLUGINS TRUE)
```
oscargm98, line 218: In my local c-blosc2 this makes the following config.h:
scrsh
Then HAVE_PLUGINS appears as undefined and blosc_init() runs the lines
  register_codecs();
  register_filters();
but it should not.

oscargm98:
Win64 tests bugs theoretically solved on https://github.com/Blosc/c-blosc2/pull/319

CMakeLists.txt
nmoinvaz: What cmake command are you using?

CMakeLists.txt
nmoinvaz: Hmm. I wonder why it wasn't registering for me. I will have to take a look at it.

oscargm98: From the Blosc Development Team, accordingly to our last meeting we agreed that the use of the set() command should be kept for consistency reasons.

CMakeLists.txt
oscargm98: I have retested it several times on my local device and it no longer gives me any error. Sorry for the inconveniences.

CMakeLists.txt
nmoinvaz: I didn't mean to cause a meeting. I wasn't aware of the config so you were right of course. I think perhaps my config header was not being read/written because I am building from a different directory.. I have added a commit which should fix that.

nmoinvaz:
Rebased.

oscargm98 (approved):

PR #323, opened 2021-06-21
Fix schunk_delete_chunk for an sframe

martaiborra:
Do not delete the chunk file when it is a special value chunk.

### DIFF
--- a/blosc/frame.c
+++ b/blosc/frame.c
@@ -3147,10 +3147,13 @@ void* frame_delete_chunk(blosc2_frame_s* frame, int nchunk, blosc2_schunk* schun
         BLOSC_TRACE_ERROR("Unable to get offset to chunk %d.", nchunk);
         return NULL;
       }
-      int err = sframe_delete_chunk(frame->urlpath, offset);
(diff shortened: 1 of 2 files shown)


blosc/frame.c
```diff
@@ -3147,10 +3147,13 @@ void* frame_delete_chunk(blosc2_frame_s* frame, int nchunk, blosc2_schunk* schun
         BLOSC_TRACE_ERROR("Unable to get offset to chunk %d.", nchunk);
         return NULL;
       }
-      int err = sframe_delete_chunk(frame->urlpath, offset);
-      if (err != 0) {
-        BLOSC_TRACE_ERROR("Unable to delete chunk!");
-        return NULL;
+      if (offset >= 0){
+        // There is no need to delete any file if it is a special chunk
```
FrancescAlted, line 3151: This comment is inside a codeblock referring to a condition that is not fulfilled there.  Could you rephrase it somehow so that it makes more sense in this context?

aleixalcacer (approved):

PR #324, opened 2021-06-21
Compile Blosc always with LZ4

oscargm98:


CMakeLists.txt
```diff
@@ -146,16 +142,14 @@ if(BUILD_LITE STREQUAL ON)
...
-    # not found, we will use the included sources for it
-    set(HAVE_LZ4 TRUE)
+if(PREFER_EXTERNAL_LZ4)
+    find_package(LZ4)
+else()
+    message(STATUS "Using LZ4 internal sources.")
 endif()
+# HAVE_LZ4 will be set to true because even if the library is
+# not found, we will use the included sources for it
+set(HAVE_LZ4 TRUE)
```
FrancescAlted, line 152: As LZ4 is set to be mandatory, it would make sense if we get rid of the HAVE_LZ4 symbol totally.

oscargm98: of course!

FrancescAlted (approved):
LGTM

PR #333, opened 2021-07-01
Documentation maintenance

oscargm98:


FrancescAlted:
Good amount of fixes!  Is that ready for merging or you prefer to wait for more to come?

oscargm98:
Better wait a little, there are some files I still would like to check.

oscargm98:
At https://github.com/Blosc/c-blosc2/blob/documentation/doc/sphinx/source/reference/index.rst
there is a reference to "63-bit data containers", is this ok or was it supposed to be "64-bit data containers"?

FrancescAlted:
> At https://github.com/Blosc/c-blosc2/blob/documentation/doc/sphinx/source/reference/index.rst
> there is a reference to "63-bit data containers", is this ok or was it supposed to be "64-bit data containers"?

This is kind of ok. It is actually 63-bit, but it is much more common to refer to this as 64-bit. So maybe using 64 is more consistent.

aleixalcacer (approved):

FrancescAlted (approved):

PR #334, opened 2021-07-02
Fix bug when updating a chunk of a cframe

martaiborra:
When updating a chunk from a cframe, the index chunk was written immediately after the updated chunk. Thus, some data was overwritten by the index chunk.


FrancescAlted (approved):
LGTM

PR #336, opened 2021-07-08
Fix leak when decompressing chunk in a frame

aleixalcacer:


FrancescAlted:
Already fixed in #335

PR #338, opened 2021-07-21
Add error message when appending a chunk which is not multiple of typesize

martaiborra:
If `chunksize` is not multiple of `typesize` that means `typesize` has not been set correctly.

### DIFF
--- a/tests/test_bitshuffle_leftovers.c
+++ b/tests/test_bitshuffle_leftovers.c
@@ -20,13 +20,10 @@ static int test_roundtrip_bitshuffle8(int size, void *data, void *data_out, void
   int isize = size;
   int osize = size + BLOSC_MIN_HEADER_LENGTH;
   int csize = blosc_compress(9, BLOSC_BITSHUFFLE, 8, isize, data, data_out, osize);
+  mu_assert("ERROR: undetected chunksize not being a multiple of typesize", csize < 0);
   if (csize == 0) {
     printf("Buffer is uncompressible.  Giving up.\n");
     return 1;
-  }
-  else if (csize < 0) {
-    printf("Compression error.  Error code: %d\n", csize);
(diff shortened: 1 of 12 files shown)


FrancescAlted (approved):

FrancescAlted (commented):
Ups, the CI is not happy for some reason.

tests/test_bitshuffle_leftovers.c
```diff
@@ -96,7 +96,7 @@ int main(void) {
    * (not supported by bitshuffle) and in addition, it is not
    * divisible by 8 (typesize) again.
    */
-  int size = 641092;
+  int size = 641096;
```
FrancescAlted, line 99: With that, size *is* divisible by 8, making the comment above untrue. After thinking twice, this module is testing the situation where the chunksize is not a multiple of typesize, and we are not supporting this anymore.  So I'd rather keep the previous value (641092) and test that `blosc_compress()` is raising an error (like `mu_assert("ERROR: undetected chunksize not being a multiple of typesize", cbytes2 < 0)`.

tests/test_compressor.c
```diff
@@ -207,7 +207,7 @@ static char *test_typesize(void) {
   mu_assert("ERROR: cbytes is not correct", cbytes < (int)size);
 
   /* Activate the BLOSC_TYPESIZE variable */
-  setenv("BLOSC_TYPESIZE", "9", 0);
+  setenv("BLOSC_TYPESIZE", "4", 0);
```
FrancescAlted, line 210: Let's keep 9 here too.

FrancescAlted (changes requested):

FrancescAlted (approved):
Great job!

FrancescAlted:
This has been overridden by #356.

PR #340, opened 2021-07-27
Update pkg-config file to support blosc2. Fixes #236.

### LINKED ISSUE #236: pkg-config blosc.pc conflict with c-blosc (v1)
I've been using c-blosc (v1) for awhile and am excited about the improvements on v2...

I just was playing around with v2 and noticed that the compiled ```blosc.pc``` pkg-config file:

1. Conflicts with the pkg-config file for c-blosc (v1)
2. Specifies the library as "blosc" when it should be "blosc2"?

I know things are still in development - if I understand the philosophy correctly, according to the main README, c-blosc2 tries to maintain API compatibility with c-blosc(v1) but it isn't 100% guaranteed. (Which is totally fine: sometimes you have to modify the API to implement features/improvements.)

Since the library is "blosc2" as is the header file for the library, shouldn't the pkg-config also be "blosc2" with appropriate references? (Maybe this is the intention all along and it just hasn't come up yet :) )

havardAasen:
Rename and update the pkg-config file.

For testing with one of the example files, one could use
```
cc simple.c `pkg-config --cflags --libs blosc2`
```

I only tested this on Debian 11, but regarding the small changes, I don't think that it should be a problem.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -410,10 +410,10 @@ endif()
 # uninstall target
 if(BLOSC_INSTALL)
     configure_file(
-         "${CMAKE_CURRENT_SOURCE_DIR}/blosc.pc.in"
-         "${CMAKE_CURRENT_BINARY_DIR}/blosc.pc"
+         "${CMAKE_CURRENT_SOURCE_DIR}/blosc2.pc.in"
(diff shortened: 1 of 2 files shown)


FrancescAlted:
Looks good to me.  Thanks!

PR #342, opened 2021-07-28
Build system: Change from hard coded 'lib' to CMAKE_INSTALL_LIBDIR. Fixes #341

### LINKED ISSUE #341: Buildsystem: Use CMAKE_INSTALL_LIBDIR instead of hard coded 'lib'
The library path is hard coded to `lib`.
Would you accept a patch to change this to the CMake variable `CMAKE_INSTALL_LIBDIR`? This is included in `GNUInstallDirs`. Some more information is located in CMake's documentation.
[https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#module:GNUInstallDirs]()
and
[https://cmake.org/cmake/help/latest/command/install.html#files]()

The variable has some nice defaults, depending on which distribution it's being used on. It's also well known and easy to change during configuration.

havardAasen:
Removes the hard coded library path `lib`. Replaces it with `CMAKE_INSTALL_LIBDIR`. This also sets the same path in the pkg-config file.

Depending on how you configure with CMake, the path will default to `lib` or the distribution specific library path. It can be overridden if specifying `-DCMAKE_INSTALL_LIBDIR=your_libdir_path` during CMake configuration. The variable is also avaliable for manual change when configuring with `ccmake`.

The `CMAKE_INSTALL_LIBDIR` is included with `GNUInstallDirs`. There is several other `CMAKE_INSTALL_*` variables like this.

The pr also includes a commit regarding some typos found by lintian.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -409,12 +409,14 @@ endif()
 
 # uninstall target
 if(BLOSC_INSTALL)
+    include(GNUInstallDirs)
+    set(CMAKE_INSTALL_LIBDIR ${CMAKE_INSTALL_LIBDIR})
     configure_file(
          "${CMAKE_CURRENT_SOURCE_DIR}/blosc2.pc.in"
          "${CMAKE_CURRENT_BINARY_DIR}/blosc2.pc"
          @ONLY)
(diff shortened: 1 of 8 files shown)


FrancescAlted:
Looks good.  Thanks!

Issue #352, opened 2021-09-08
SIGSEGV error on sframe_bench


FrancescAlted:
Here it is my output:

```
faltet@Mac-mini-de-Francesc ~/b/c/build (main)> ./bench/sframe_bench                                                                                          (base)
*******************************************************
***** Creating the frame and sframe with 1000 chunks ****
*******************************************************
Test comparation frame vs sframe with 1000 chunks.
[Sframe Compr] Elapsed time:	  0.636 s.  Processed data: 0.745 GB (1.171 GB/s)
[Cframe Compr] Elapsed time:	  0.492 s.  Processed data: 0.745 GB (1.514 GB/s)
Compression super-chunk-sframe: 800000000 -> 4010025 (199.5x)
Compression super-chunk-cframe: 800000000 -> 4010025 (199.5x)
[Sframe Decompr] Elapsed time:	  0.094 s.  Processed data: 0.745 GB (7.906 GB/s)
[Cframe Decompr] Elapsed time:	  0.092 s.  Processed data: 0.745 GB (8.065 GB/s)
Decompression successful!
Successful roundtrip!
fish: Job 1, './bench/sframe_bench' terminated by signal SIGSEGV (Address boundary error)
```

FrancescAlted:
Fixed in #353.

PR #353, opened 2021-09-08
fix bug in sframe_bench.c

martaiborra:
Fix bug when running the sframe_bench with no operation parameter.
This should fix the [issue](https://github.com/Blosc/c-blosc2/issues/352#issue-990807263) reported.


FrancescAlted:
Great.  Thanks!

PR #355, opened 2021-09-16
add blosc2_rename_urlpath function

martaiborra:
Add a portable function to rename a file or a full directory.

### DIFF
--- a/blosc/directories.c
+++ b/blosc/directories.c
@@ -18,8 +18,8 @@
   #include <io.h>
 
   int blosc2_remove_dir(const char* dir_path) {
-    char last_char = dir_path[strlen(dir_path) - 1];
     char* path;
+    char last_char = dir_path[strlen(dir_path) - 1];
     if (last_char != '\\' || last_char != '/') {
       path = malloc(strlen(dir_path) + 2 + 1);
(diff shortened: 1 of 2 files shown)


blosc/directories.c
```diff
@@ -49,86 +67,220 @@
...
+      if (strcmp(".", cfile.name) == 0 || strcmp("..", cfile.name) == 0) {
+        continue;
+      }
+      fname = malloc(strlen(path) + strlen(cfile.name) + 1);
+      sprintf(fname, "%s%s", path, cfile.name);
+
+      new_fname = malloc(strlen(new_path) + strlen(cfile.name) + 1);
+      sprintf(new_fname, "%s%s", new_path, cfile.name);
+
+      ret = rename(fname, new_fname);
```
FrancescAlted, line 125: I think rename accepts dir entries besides files (at least on UNIX).  That would be single operation for the whole directory instead of many.  Can you have a try at it?

FrancescAlted (commented):
Could you address my comment?

blosc/directories.c
martaiborra: You were right, even on Windows.

FrancescAlted (approved):
Looks good!

Issue #351, opened 2021-09-07
Bad roundtrip error


FrancescAlted:
This error triggers when you use a large enough number of rows in test_update_chunk.c:

```
diff --git a/tests/test_update_chunk.c b/tests/test_update_chunk.c
index 38cb4da0..1cd8d212 100644
--- a/tests/test_update_chunk.c
+++ b/tests/test_update_chunk.c
@@ -30,11 +30,7 @@ typedef struct {
 } test_ndata;

 test_ndata tndata[] = {
-    {1, 4},
-    {10, 4},
-    {5,  0},
-    {33, 32},
-    {1,  0}
+    {25000, 0},
 };

 typedef struct {
```

The error is:

```
> ./tests/test_update_chunk                                                                                       (base)
F (ERROR: bad roundtrip)
	Tests run: 1
```

However, if instead of 25000 one puts 20000, this works.

FrancescAlted:
Fixed in #361

PR #361, opened 2021-10-04
Fix test updates and inserts for 64-bit schunks

FrancescAlted:
This PR fixes some tests so as to support super-chunks larger than 32-bit, as stated in ticket #351.


FrancescAlted:
This fixes #351.

PR #362, opened 2021-11-02
Fix build on FreeBSD/powerpc64le

pkubaj:
64 bit POWER little endian is called powerpc64le on FreeBSD.


FrancescAlted:
LGTM.  Thanks!

PR #363, opened 2021-11-04
Remove to force including plugins sources into tests

oscargm98:


plugins/codecs/ndlz/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ set(SOURCES ${SOURCES} ndlz.c ndlz.h ndlz-private.h ndlz4x4.c ndlz4x4.h ndlz8x8.
```
FrancescAlted, line 3: You don't need `set(SOURCES...)` line anymore, right?

FrancescAlted (changes requested):

plugins/codecs/ndlz/CMakeLists.txt
FrancescAlted: And the same goes for other cmake files.

oscargm98: Then no plugin source will be distributed. If we want plugins to be in the distribution library, then they must be into SOURCES first.

FrancescAlted: Oh, my mistake.  So if I understand correctly, the updated `$SOURCES` variable is used *outside* of this cmake file, right?  If so, I have no objections.

oscargm98: That is what I think. This evening I make sure of it and I confirm you.

PR #368, opened 2021-12-01
nmetalayers is unsigned, it cannot be negative

DimitriPapadopoulos:
https://github.com/Blosc/c-blosc2/blob/b2434ad1be6ea46e101e0691bfd6b2a43174cc8a/blosc/frame.c#L1238


FrancescAlted:
Yes, well spotted.

PR #369, opened 2021-12-01
typesize is unisgned, it cannot be negative

DimitriPapadopoulos:
https://github.com/Blosc/c-blosc2/blob/b2434ad1be6ea46e101e0691bfd6b2a43174cc8a/blosc/blosc2.c#L622

Additionally, the value of both `BLOSC_MAX_TYPESIZE` and `UINT8_MAX` is 255. It doesn't make sense to test whether an `uint8_t` is larger than 255. Perhaps it would make sense to define `BLOSC_MAX_TYPESIZE` as `UINT8_MAX` instead?
https://github.com/Blosc/c-blosc2/blob/b2434ad1be6ea46e101e0691bfd6b2a43174cc8a/include/blosc2.h#L105-L106


FrancescAlted:
Yep.

DimitriPapadopoulos:
See also #371 which modifies the error message after f553bd2.

PR #370, opened 2021-12-01
Correct test variable.

havardAasen:
Making sure that the test doesn't check the value against itself.


FrancescAlted:
Well spotted.

PR #375, opened 2022-01-21
create get_offsets function

martaiborra:
This function can only be used when the schunk has a frame. It returns the decompressed offsets chunk.


FrancescAlted (approved):
Looks good to me

PR #376, opened 2022-02-04
fix: surround all headers in blosc2.h with extern "C" in C++ mode

theHamsta:
Problem: not all headers in `blosc.h` are surrounded with a `extern "C"` in C++ mode. This leads to linkage problems with using this lib from C++.

This problem can be solved either by surrounding the `blosc2-*` headers in the top-level header (if only this header is intended for C++ use without manual `extern "C"`) or by using `extern "C"` in all headers. Please tell me which solution is preferred!

### DIFF
--- a/include/blosc2.h
+++ b/include/blosc2.h
@@ -25,9 +25,16 @@
 #include <stdint.h>
 #include <stdbool.h>
 #include <stdio.h>
+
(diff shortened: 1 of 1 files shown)


FrancescAlted:
Your fix looks good to me. Will merge if all tests pass. Thanks!

PR #378, opened 2022-02-18
Allow using clang on ppc64le

pkubaj:
Clang 14 also has SSE translation support.


FrancescAlted:
Great.  Can you point to some documents on how this SSE translation works in Clang 14?

PR #379, opened 2022-02-19
Correct ppc64le condition

pkubaj:
MATCHES should be used to allow for a regular expression.
STREQUAL was using the expression as a literal string.


FrancescAlted:
Thanks!

Issue #381, opened 2022-03-07
Add a function to list all the metalayers


martaiborra:
It would be convenient to add a function to list all the vlmetalayers, and another one to list all the metalayers.

PR #382, opened 2022-03-22
add vlmeta_get_names

martaiborra:
This PR addresses #381

### DIFF
--- a/include/blosc2.h
+++ b/include/blosc2.h
@@ -1910,9 +1910,13 @@ BLOSC_EXPORT int blosc2_vlmeta_get(blosc2_schunk *schunk, const char *name,
 BLOSC_EXPORT int blosc2_vlmeta_delete(blosc2_schunk *schunk, const char *name);
 
 /**
- * @brief Get a list of all the variable-length metalayer names
+ * @brief Get a list of all the variable-length metalayer names.
  * @param schunk The super-chunk containing the variable-length metalayers.
  * @param names The pointer to a char** to store the name pointers.
+ *
+ * @warning The @p **names receives a malloc'ed buffer of size @p *schunk->nvlmetalayers * sizeof(char*).
+ * The user is responsible of freeing it.
+ *
  * @return If successful, the number of the variable-length metalayers in the super-chunk.
  */
 BLOSC_EXPORT int blosc2_vlmeta_get_names(blosc2_schunk *schunk, char **names);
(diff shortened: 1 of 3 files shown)


include/blosc2.h
```diff
@@ -1909,6 +1909,14 @@ BLOSC_EXPORT int blosc2_vlmeta_get(blosc2_schunk *schunk, const char *name,
  */
 BLOSC_EXPORT int blosc2_vlmeta_delete(blosc2_schunk *schunk, const char *name);
 
+/**
+ * @brief Get a list of all the variable-length metalayer names
```
FrancescAlted, line 1913: Missing end dot.

include/blosc2.h
```diff
@@ -1909,6 +1909,14 @@ BLOSC_EXPORT int blosc2_vlmeta_get(blosc2_schunk *schunk, const char *name,
...
+ * @param schunk The super-chunk containing the variable-length metalayers.
+ * @param names The pointer to a char** to store the name pointers.
+ * @return If successful, the number of the variable-length metalayers in the super-chunk.
```
FrancescAlted, line 1916: Add an empty line before the `@return` docstring.

FrancescAlted: Also, it would be convenient to to add a note saying that the user should `free(*names)` to avoid leaks.

tests/test_schunk.c
```diff
@@ -53,10 +53,27 @@ static char* test_schunk(void) {
 
   blosc2_meta_update(schunk, "metalayer2", (uint8_t *) "my metalayer2", sizeof("my metalayer2"));
 
+  char names_[2][13] = {"vlmetalayer1", "vlmetalayer2"};
+  char **names = malloc(sizeof(char**));
```
FrancescAlted, line 57: I *think* you need to know the number of metalayers (`nvlmeta`) before the malloc (which should be `malloc(nvlmeta * sizeof(char*))`).

martaiborra: Fixed in 52736067b453102049a645aa7137388b5a23e4be.

include/blosc2.h
martaiborra: Fixed in 52736067b453102049a645aa7137388b5a23e4be.

include/blosc2.h
martaiborra: Fixed in 52736067b453102049a645aa7137388b5a23e4be.

PR #385, opened 2022-04-05
Fix warnings and change nchunks to int64_t

martaiborra:


FrancescAlted (approved):
Looks good to me.  There is still some warnings on pretty inoffensive formatting strings (apparently '%lld' is not the same than a `int64_t`), but that should wait for the next round.

PR #390, opened 2022-04-07
Fix cutest warnings

aleixalcacer:


FrancescAlted:
Now gcc-10 is clean of warnings, so :+1:

PR #393, opened 2022-04-10
WIP: Allow users to use an external zlib-ng

hmaarrfk:
This supersedes the previous PR that allowed users to use an external ZLIB-NG.

I think it works because it is less invasive.

- [x] Works with internal zlib-ng
- [x] Works with external zlib-ng
- [x] Works with external zlib
- [x] Graceful fallback to built in zlib


blosc/blosc2.c
```diff
@@ -43,7 +43,11 @@
   #include <ippdc.h>
 #endif
 #if defined(HAVE_ZLIB_NG)
+#ifdef ZLIB_COMPAT
   #include "zlib.h"
+#else
+  #include "zlib-ng.h"
+#endif
```
FrancescAlted, line 50: By selecting "zlib-ng.h" you are introducing yet another format to Blosc2, which is not what we'd like to do.  We have chosen zlib-ng as a modern implementation of zlib, but that was always meant to be run in ZLIB_COMPAT mode.  What we want is to use the external zlib-ng, but in ZLIB_COMPAT mode.  Is that possible?

FrancescAlted (changes requested):

FrancescAlted:
Oops, it looks like I was wrong and zlib-ng is meant to *always* produce bit streams that are compatible with the original zlib project: https://github.com/zlib-ng/zlib-ng/discussions/905. Sorry about the confusion.

With that, I am happy with this PR. If it is this ready to be merged, I will gladly do so :-)

blosc/blosc2.c
```diff
@@ -3348,7 +3364,13 @@ int blosc_get_complib_info(const char* compname, char** complib, char** version)
   }
 #if defined(HAVE_ZLIB)
   else if (clibcode == BLOSC_ZLIB_LIB) {
+#ifdef ZLIB_COMPAT
     clibversion = ZLIB_VERSION;
+#elif defined(HAVE_ZLIB_NG)
+    clibversion = ZLIBNG_VERSION;
```
hmaarrfk, line 3370: Does the return value of clib version get saved in the file? 

Zlibng version will be in the 2.x.x while zlib version. 1.x.x which may be an important change

hmaarrfk: And potentially b undesirable

FrancescAlted: That's a good question.  Maybe @nmoinvaz can shed some light here.

FrancescAlted: Ups, sorry, `clibversion` is only used for `blosc_get_complib_info()` to get the versions of internal codecs, so not stored in the file.  With that, I think we are safe merging that.

FrancescAlted (approved):

FrancescAlted:
All done.  Thanks @hmaarrfk !

hmaarrfk:
Thanks for this!

hmaarrfk:
https://github.com/Blosc/c-blosc2/pull/358 might also provide insights as to why things aren't working here

PR #397, opened 2022-05-03
.qco Codec (scratch)

mwlon:


FrancescAlted:
Hi @mwlon .  Thanks for your contribution!  However, and despite that the QC codec would be interesting indeed for C-Blosc2, we cannot accept external dependencies that are in languages that are different than C; this requisite is expressed here: https://github.com/Blosc/c-blosc2/blob/main/plugins/README.md#requirements-for-registering-plugins.
In case a pure C version is encoded in the future, we would be glad to reconsider this.  Meanwhile my advise is that you use QC as a local registered codec.

PR #400, opened 2022-05-07
Add CI for ARM SF/HF and AARCH64

nmoinvaz:
Running the CMake tests on ARM CI takes about 10 minutes longer than non-emulated.


nmoinvaz:
With this same setup you should be able to test any of the architectures that zlib-ng does, you would just need to add the items to the CI matrix and then copy over the cmake toolchain file.

FrancescAlted (approved):
Very nice.  Thanks!

PR #402, opened 2022-05-16
Use proper types for native zlib-ng interface

zklaus:
This changes the types for the zlib wrappers in case of the native zlib-ng to follow the upstream interface.
Not only does this save some casting, it also makes sure that the types are fitting. Without this, in some cases the incoming `size_t` may not be compatible with the `uLongf` used by the original zlib, for example when MSVC uses the LLP64 model on windows.

This surfaced in conda-forge/c-blosc2-feedstock#22


FrancescAlted:
That looks great.  Thanks for your contribution!

PR #403, opened 2022-06-02
Function index_unidim_to_multidim() moved to plugin_utils

oscargm98:


plugins/plugin_utils.c
```diff
@@ -97,6 +98,19 @@ int32_t deserialize_meta(uint8_t *smeta, int32_t smeta_len, int8_t *ndim, int64_
   return slen;
 }
 
+void index_unidim_to_multidim(uint8_t ndim, int64_t *shape, int64_t i, int64_t *index) {
+    int64_t strides[8];
```
FrancescAlted, line 102: This 8 could be a constant, like `BLOSC2_MAXDIM` in blosc2.h.

plugins/plugin_utils.h
```diff
@@ -9,4 +9,6 @@ void swap_store(void *dest, const void *pa, int size);
 int32_t deserialize_meta(uint8_t *smeta, int32_t smeta_len, int8_t *ndim, int64_t *shape,
                          int32_t *chunkshape, int32_t *blockshape);
 
+void index_unidim_to_multidim(uint8_t ndim, int64_t *shape, int64_t i, int64_t *index);
+
 void index_multidim_to_unidim(const int64_t *index, int8_t ndim, const int64_t *strides, int64_t *i);
```
FrancescAlted, line 14: If we want `index_unidim_to_multidim ` and `index_unidim_to_multidim` to be public, I propose to rename them to `blosc2_unidim_to_multidim` and `blosc2_unidim_to_multidim` and add them to the end of blosc2.h

plugins/plugin_utils.c
oscargm98: no m'apareix cap constant BLOSC2_MAXDIM

plugins/plugin_utils.h
oscargm98: It makes sense, I will put it on blosc2.h

plugins/plugin_utils.c
FrancescAlted: Yeah, this is why I am suggesting to *add* it :-)

FrancescAlted (approved):
LGTM

PR #407, opened 2022-06-28
Fix bug in update_chunk

martaiborra:
If a special chunk is added to a schunk, the chunk file is not created leaving its index for another non special value chunk. But when updating it by a non special value chunk, a new index must be found in order to not overwrite an existing chunk.


FrancescAlted (approved):
I assume that the test modification is checking the buggy case.  Other than that, LGTM.

PR #409, opened 2022-07-01
Add file_offset functionality

oscargm98:
* New parameter file_offset of the blosc2_frame_s struct
* New blosc2_schunk_open_offset() function
* New blosc2_schunk_append_file() function
* New test_frame_offset.c test


blosc/frame.c
```diff
@@ -847,6 +850,88 @@ blosc2_frame_s* frame_from_file(const char* urlpath, const blosc2_io *io) {
 }
 
 
+/* Initialize a frame out of a file */
+blosc2_frame_s* frame_from_file_offset(const char* urlpath, const blosc2_io *io, int64_t offset) {
```
FrancescAlted, line 854: It would be nice if you can remove the `frame_from_file()` function, and replace it by calls to `frame_from_file_offset` with the `offset` param set to 0.

FrancescAlted (changes requested):
I have a comment.  But other than this, everything looks good!

blosc/frame.c
oscargm98: Done!

FrancescAlted (approved):
LGTM

PR #414, opened 2022-08-18
New blosc1 api

FrancescAlted:
API change allowing to compile apps with both C-Blosc(1) and C-Blosc2.


include/blosc2.h
```diff
@@ -48,6 +48,40 @@ extern "C" {
 extern "C" {
 #endif
 
+// For compatibility with the Blosc 1.x series
+#ifdef BLOSC1_COMPAT
+  // Blosc2 symbols that should be accessible from Blosc 1.x API
+  #define BLOSC2_MAX_OVERHEAD BLOSC_MAX_OVERHEAD
+  #define BLOSC2_MAX_BUFFERSIZE BLOSC_MAX_BUFFERSIZE
```
MehdiChinoune, line 60: Shouldn't be the opposite?
```CXX
#define BLOSC_MAX_OVERHEAD BLOSC2_MAX_OVERHEAD
```

FrancescAlted: Good point!  Fixed in 5c54e545

Thanks

PR #415, opened 2022-09-02
Updating navigation to be consistent with new website

datapythonista:
Update the docs so the navigation bar shows the same options as the website, the logo link points to the global home...


FrancescAlted (approved):
LGTM

PR #416, opened 2022-09-02
Simplifying and standardizing directory structure

datapythonista:
Making the directory structure follow the same convention as python-blosc and python-blosc2 (using simply `doc/` instead of `doc/sphinx/source`)


FrancescAlted:
LGTM

PR #425, opened 2022-10-20
blosc/blosc2.c: optimize the mutex lock with spin lock

jiebinn:
The mutex lock cost lots of cycles in t_blosc_do_job when the thread number is large. The updating to global values of protected part is not that heavy. It will save lots of cycles if we replace the mutex lock with spin lock.

Apply the patch and test the c-blosc2/bench.
$ ./bench/b2bench blosclz noshuffle single 160 83886080 4

Score gain: 7.4%
CPU: ICX 8380 x 2 sockets
Core number: 160 threads

### DIFF
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -16,6 +16,7 @@
 #include <sys/types.h>
 #include <assert.h>
 #include <math.h>
+#include <pthread.h>
 
 #include "blosc2.h"
 #include "blosc-private.h"
@@ -3055,10 +3056,10 @@ static void t_blosc_do_job(void *ctxt)
   }
   else {
     // Use dynamic schedule via a queue.  Get the next block.
-    pthread_mutex_lock(&context->count_mutex);
+    pthread_spin_lock(&context->count_spin);
     context->thread_nblock++;
     nblock_ = context->thread_nblock;
-    pthread_mutex_unlock(&context->count_mutex);
+    pthread_spin_unlock(&context->count_spin);
     tblock = nblocks;
   }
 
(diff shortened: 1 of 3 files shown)


jiebinn:
> That's nice, but 160 threads looks like too much to me. Which scenarios can benefit from such a large number of threads? Do you experience speed-ups for less number of threads (say <= 64)?
> 
> Also, it looks like the spin lock is only available on Linux (i.e. Mac and Win tests are failing).

Hi Francesc, 
     I have got the performance test result. It has got 3.5% or 1.9% performance improvement when the thread number is 64 or 32. 
     For win, I think it could fall back to the original path ("EnterCriticalSection"). 
     For Mac, do you have any good idea? Or should we use pthread_rwlock_wrlock instead in Mac?

FrancescAlted:
Thanks.  So for reasonable amounts of threads (~ 32) the gain is pretty small (1.9%).  In my experience, you never get much speedup in compression/decompression when using more than say, 32 threads (even if the number of cores is larger).  Another thing is that you are using a filter or codec that are very CPU demanding (e.g. Zstd with clevel 9); but in this case I doubt that this patch would buy us a lot of performance.

Another thing that I don't quite understand is that you keep using blosclz without any filter (nofilter).  When using Blosc the normal scenario is to use it for compressing numerical binary data, which normally benefits (sometimes a lot) when using shuffle or bitshuffle.  Are you seeing any improvement in these cases too?

jiebinn:
> Thanks. So for reasonable amounts of threads (~ 32) the gain is pretty small (1.9%). In my experience, you never get much speedup in compression/decompression when using more than say, 32 threads (even if the number of cores is larger). Another thing is that you are using a filter or codec that are very CPU demanding (e.g. Zstd with clevel 9); but in this case I doubt that this patch would buy us a lot of performance.
> 
> Another thing that I don't quite understand is that you keep using blosclz without any filter (nofilter). When using Blosc the normal scenario is to use it for compressing numerical binary data, which normally benefits (sometimes a lot) when using shuffle or bitshuffle. Are you seeing any improvement in these cases too?

Hi Francesc,
     I have detected the issue by running the Phoronix Test Suite (pts/blosc-1.2.0). That's why we have used the default configuration "blosclz noshuffle". The default CPU_NUM in PTS is the maximum CPU threads in current system (160 on ICX server). The test scenario is to make the performance of benchmark/workload much more scalable as the CPU/threads number increases. The test platform is ICX or latest server platform.
     BTW, I have got the performance data if we use shuffle together with blosclz and 32 threads. The performance gains 4.5%.
     $ ./bench/b2bench blosclz shuffle single 32 83886080 4

FrancescAlted:
As this has been quiet for some time, there are still pending issues with Windows and Mac, and the performance improvements are rather meager, I think it is better to close this one (we are in Closember after all).

PR #428, opened 2022-12-06
Change user defined filters signature

martaiborra:
This is needed for supporting user defined filters in python-blosc2.


FrancescAlted (approved):
LGTM

PR #431, opened 2022-12-14
Add support for macos `universal2` (`arm64`+`x86_64` build)

t20100:
This PR proposes a fix to allow building c-blosc2 as part of a Python package that is built as a universal2 binary package on macos with enabled AVX2/SSE2 optimisation for the x86_64 target only, NEON for the arm target and ALTIVEC for ppc64le.

To achieve this, c-blosc2 should support defining SHUFFLE_*_ENABLED and be able to build all files (including *-avx2.c, *-sse2.c, *-neon.c and *-altivec.c) even when the compilation target does not support the given instruction set.

This is done by toggling SIMD instructions usage with both SHUFFLE_*_ENABLED and __AVX2__, __SSE2__, __ARM_NEON and __ALTIVEC__.

closes #430

related to https://github.com/Blosc/c-blosc/pull/347
related to https://github.com/silx-kit/hdf5plugin/pull/201

PR #436, opened 2023-01-09
Fixing compilation with optimisation: Set optimisation flags when building shuffle.c

t20100:
This PR fixes a compilation issue introduced in PR #431 (commit https://github.com/Blosc/c-blosc2/commit/c6031717b53942ed221df361998c46c5eade2dae) and first raised here: https://github.com/Blosc/c-blosc2/pull/431#issuecomment-1375586510.

The definition of `SHUFFLE_USE_*` made in `shuffle.h` relies on hardware optimisation macros (`__SSE2__`, `__AVX2__`, `__ARM_NEON` and `__ALTIVEC__`), so the hardware optimisation compilation flags must also be set where `SHUFFLE_USE_*` macros are used, thus in `shuffle.c`.

`blosc2.c` also includes `shuffle.h`, but does not use `SHUFFLE_USE_*` macros. Should it also be compiled with the optimisations flags? or all blosc source to avoid such issues in the future?

### DIFF
--- a/blosc/CMakeLists.txt
+++ b/blosc/CMakeLists.txt
@@ -172,11 +172,17 @@ if(COMPILER_SUPPORT_SSE2)
             set_source_files_properties(
                     shuffle-sse2.c bitshuffle-sse2.c blosclz.c fastcopy.c
                     PROPERTIES COMPILE_FLAGS "/arch:SSE2")
+            set_property(
+                    SOURCE shuffle.c
+                    APPEND PROPERTY COMPILE_OPTIONS "/arch:SSE2")
         endif()
     else()
         set_source_files_properties(
                 shuffle-sse2.c bitshuffle-sse2.c blosclz.c fastcopy.c
                 PROPERTIES COMPILE_FLAGS -msse2)
+        set_property(
+                SOURCE shuffle.c
+                APPEND PROPERTY COMPILE_OPTIONS -msse2)
     endif()
 
     # Define a symbol for the shuffle-dispatch implementation
@@ -191,10 +197,16 @@ if(COMPILER_SUPPORT_AVX2)
         set_source_files_properties(
                 shuffle-avx2.c bitshuffle-avx2.c
                 PROPERTIES COMPILE_FLAGS "/arch:AVX2")
+        set_property(
+                SOURCE shuffle.c
+                APPEND PROPERTY COMPILE_OPTIONS "/arch:AVX2")
     else()
         set_source_files_properties(
                 shuffle-avx2.c bitshuffle-avx2.c
                 PROPERTIES COMPILE_FLAGS -mavx2)
+        set_property(
+                SOURCE shuffle.c
+                APPEND PROPERTY COMPILE_OPTIONS -mavx2)
     endif()
 
     # Define a symbol for the shuffle-dispatch implementation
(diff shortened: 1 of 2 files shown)


blosc/CMakeLists.txt
```diff
@@ -172,11 +172,17 @@ if(COMPILER_SUPPORT_SSE2)
             set_source_files_properties(
                     shuffle-sse2.c bitshuffle-sse2.c blosclz.c fastcopy.c
                     PROPERTIES COMPILE_FLAGS "/arch:SSE2")
+            set_property(
+                    SOURCE shuffle.c
+                    APPEND PROPERTY COMPILE_OPTIONS "/arch:SSE2")
```
t20100, line 177: Using [`COMPILE_OPTIONS`](https://cmake.org/cmake/help/latest/prop_tgt/COMPILE_OPTIONS.html) here since [`COMPILE_FLAGS`](https://cmake.org/cmake/help/latest/prop_tgt/COMPILE_FLAGS.html) is deprecated (see https://cmake.org/cmake/help/latest/prop_tgt/COMPILE_FLAGS.html).

Also `COMPILE_OPTIONS` is a `;`-separated list which fits `set_property` as opposed to `COMPILE_FLAGS`

FrancescAlted:
I tested this, but it does not work for me:

image

Do you have access to the perf tool on Linux?  It's a nice way to check that AVX2 is being used.

t20100:
I also reverted commit [89f9335](https://github.com/Blosc/c-blosc2/pull/431/commits/89f93350b5fabde43ed37974fef817d9ee26da98). This breaks the `universal2` build, but it seems it is not the way to fix it. I'm looking for a way to make it work.

With this, I get the following for `perf record -e cpu-cycles ./bench/b2bench lz4 bitshuffle single`: 

image

FrancescAlted:
Yes, your last fix seems to work for me too.  I just have a comment.  Thanks!

blosc/CMakeLists.txt
t20100: Yes, it is needed or it needs other modifications:

`shuffle.c` is using `SHUFFLE_USE_AVX2` which is defined as:
 https://github.com/Blosc/c-blosc2/blob/e204359a202ce8624eed47f56bb2093dab43ce60/blosc/shuffle.h#L28-L30
So the definition needs `-mavx2` flag to have `__AVX2__` defined. Same applies for SSE2/NEON/ALTIVEC.

To avoid this, it looks to require to change the conditional include of `[bit]shuffle-*.h` and the way `get_shuffle_implementation` is implemented.

t20100:
I haven't checked yet the build for `universal2` and I expect it too fail.

blosc/CMakeLists.txt
FrancescAlted: I am fine with that.  I just want to make sure we don't abuse of SIMD flags.

FrancescAlted:
> I haven't checked yet the build for `universal2` and I expect it too fail.

It would be possible to disable AVX2 just for `universal` builds?  That may provide an easy workaround.

t20100:
I checked `universal2` and it works fine, ` && !defined(__IMMINTRIN_H)` is doing the work also in this case.

I just updated the comment at the end of `#if...#end` in `shuffle.c`: https://github.com/Blosc/c-blosc2/pull/436/files#diff-a10ebc109ea502babcc7b8845fc8c74137389e4a46d30667e2bfee427f21e020R163

FrancescAlted:
Thanks @t20100 !

PR #438, opened 2023-01-11
Merge caterva

FrancescAlted:
This is meant to merge the Caterva library into C-Blosc2


DimitriPapadopoulos:
Perhaps it would have been better not to reformat vendored files [`plugins/codecs/ndlz/xxhash.h`](https://github.com/Blosc/c-blosc2/blob/main/plugins/codecs/ndlz/xxhash.h) and [`plugins/codecs/ndlz/xxhash.c`](https://github.com/Blosc/c-blosc2/blob/main/plugins/codecs/ndlz/xxhash.c).

PR #445, opened 2023-02-06
Data type is part of the b2nd metalayer now

FrancescAlted:
This PR adds support for a data type in the b2nd metalayer.  This allows to specify a full-fledged NumPy dtype in b2nd arrays.  If `dtype` is passed as NULL in the context constructor, the default is to set it as "uint8" (i.e. plain bytes) internally.

Note: for compatibility with the (now deprecated) caterva metalayer, when a `dtype` entry is not found there, dtype is automatically set to "uint8".


README_B2ND_METALAYER.rst
```diff
@@ -51,3 +51,13 @@ Finally, the `blockshape` section is meant to store the actual block shape info:
       |                |                      +--[msgpack] int32
       |                +--[msgpack] int32
       +--[msgpack] int32
+
+Finally, the `dtype` section is meant to store the data type information.  Its representation follows the NumPy
+convention (as in `repr(np.dtype)`; e.g. an `int32_t` dtype is represented as "int32")::
```
martaiborra, line 56: The `repr(np.dtype("int32"))` returns "dtype('int32')". It would be easier to rebuild the dtype in Python if we stored `str(a)`, which would give 'int32'. Another option is to use `np.dtype('int32').str`, which would give `<i4`.

martaiborra (approved):
Other than the comment above, this looks good to me.

README_B2ND_METALAYER.rst
FrancescAlted: Good call.  Fixed in a5b1df9b.

FrancescAlted: BTW, I don't like `np.dtype('int32').str` because `np.dtype("u1,i8").str` gives `'|V9'`, whereas `str(np.dtype("u1,i8"))` returns the more meaningful `"[('f0', 'u1'), ('f1', '<i8')]"`.  So we stay with `str(np.dtype)`.

PR #448, opened 2023-02-17
export more schunk API

bnavigator:
Fixes https://github.com/Blosc/python-blosc2/issues/83

These are documented in the API docs: https://www.blosc.org/c-blosc2/reference/schunk.html, why should they not be exported?


FrancescAlted:
LGTM

PR #449, opened 2023-02-18
zstd sources updated to 1.5.4

DimitriPapadopoulos:
I have extracted the `lib` directory from the Zstandard tarball, and discarded the rest, as was the case for previous versions.


FrancescAlted:
LGTM

PR #452, opened 2023-02-26
b2nd: Use const where possible.

cf-natali:
The current API doesn't use `const` on pointers when treating them as read-only, which makes the API difficult to use in a const-safe way. It also makes it difficult for users to understand the semantics.

Note that casts had to be used internally in a few places because of the use of functions which handle both read and write paths depending on a flag.

### DIFF
--- a/blosc/b2nd.c
+++ b/blosc/b2nd.c
@@ -17,8 +17,8 @@
 #include <inttypes.h>
 
 
-int b2nd_serialize_meta(int8_t ndim, int64_t *shape, int32_t *chunkshape,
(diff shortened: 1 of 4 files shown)


cf-natali:
Note that there's probably a few other functions/parameters which can be updated - I'd like to first get a feel on whether that's an acceptable change before doing more work on this.

FrancescAlted:
Thanks for this @cf-natali !  I like what you are doing.  Tell me when are you are done and will be happy to merge :-)

FrancescAlted:
Let's merge this.  @cf-natali in case you have more const work pending, feel free to open another PR.  Thanks!

PR #454, opened 2023-03-02
Optimise PNG compression

DimitriPapadopoulos:
Use `optipng` to reduce size of PNG files up to 50%.


FrancescAlted:
Thanks @DimitriPapadopoulos !

PR #456, opened 2023-03-08
Support for bytedelta filter

FrancescAlted:
This bytedelta filter makes a delta for every byte on a data stream.  This is implemented as plugin for Blosc2, and mainly meant to be used after a shuffle filter.

This is essentially based on @aras-p fascinating blog: https://aras-p.info/blog/2023/03/01/Float-Compression-7-More-Filtering-Optimization/, but removing the shuffle part. For speed, it would make sense to intertwine shuffle and bytedelta, as it is done in the blog, but that would require far more work, as Blosc supports shuffle for general (2 to 255) typesizes (channels in the blog jargon).

This is still a bit preliminary (I still need to assess the new dependency on SSE4.1). Some benchmarks should follow soon too.


blosc/CMakeLists.txt
```diff
@@ -185,13 +185,22 @@ if(COMPILER_SUPPORT_SSE2)
...
         set_source_files_properties(
                 shuffle-sse2.c bitshuffle-sse2.c blosclz.c fastcopy.c
                 PROPERTIES COMPILE_FLAGS -msse2)
         set_property(
                 SOURCE shuffle.c
                 APPEND PROPERTY COMPILE_OPTIONS -msse2)
+        # Add SIMD flags for the bytedelta filter and Intel (it seems that ARM64 does not need these)
+        set_source_files_properties(
+                ${PROJECT_SOURCE_DIR}/plugins/filters/bytedelta/bytedelta.c
+                PROPERTIES COMPILE_OPTIONS "-msse4.2")
```
aras-p, line 203: I think code as-is does not actually need SSE4.2 -- just SSSE3 should be enough (due to `_mm_shuffle_epi8` and `_mm_alignr_epi8` which are both SSSE3). So clang/gcc should be fine with just `-mssse3` flag, and MSVC does not actually need any special flag.

FrancescAlted: You are right indeed.  Thanks!

martaiborra (commented):
LGTM!

PR #461, opened 2023-03-28
Plugins improvements

oscargm98:
* Disable SHUFFLE filter when using ZFP
* Use macros for plugins tests
* Making blocks bigger than cells for example_item_prices.b2nd


plugins/codecs/ndlz/test_ndlz.c
```diff
@@ -71,17 +71,17 @@ static int test_ndlz_4(blosc2_schunk *schunk) {
 
     decompressed = blosc2_schunk_decompress_chunk(schunk, ci, data_in, chunksize);
     if (decompressed < 0) {
-      printf("Error decompressing chunk \n");
-      return -1;
+      BLOSC_TRACE_ERROR("Error decompressing chunk \n");
+      return BLOSC2_ERROR_FAILURE;
```
FrancescAlted, line 75: It is not necessary to use the `BLOSC2_ERROR_FAILURE  / BLOSC_TRACE_ERROR ` mechanism in tests or executables in general.  This is more meant for the library itself, because it is designed to be used from other programs, and messing with standard I/O is not a good practice.

FrancescAlted (changes requested):
This still needs some work.  See my comments.

PR #470, opened 2023-04-16
Consistent indentation

DimitriPapadopoulos:


include/blosc2.h
```diff
@@ -37,7 +37,7 @@ extern "C" {
 #endif
 
 #if defined(_WIN32) && !defined(__MINGW32__)
-#include <windows.h>
+  #include <windows.h>
   #include <malloc.h>
```
DimitriPapadopoulos, line 41: Perhaps it's best **not** to indent here, isn't it?

FrancescAlted: Yeah, perhaps you are right.  Could you please provide a patch with no indents for `#include` statements?

DimitriPapadopoulos: Done

PR #488, opened 2023-05-14
Fix GCC warning: ignoring return value of `fgets`

DimitriPapadopoulos:
GCC 11.3.0 on Ubuntu 22.04:
```
/my/path/c-blosc2/include/blosc2/plugins-utils.h:44:3: warning: ignoring return value of ‘fgets’ declared with attribute ‘warn_unused_result’ [-Wunused-result]
   44 |   fgets(python_path, PATH_MAX, fp);
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/include/blosc2/plugins-utils.h
+++ b/include/blosc2/plugins-utils.h
@@ -41,7 +41,7 @@ static inline void* load_lib(char *plugin_name, char *path) {
   char python_path[PATH_MAX] = {0};
   FILE *fp = popen("python -c \"exec(\\\"import sys\\npaths=sys.path\\nfor p in paths:\\n\\tif 'site-packages' in p:"
                    "\\n \\t\\tprint(p+'/', end='')\\n \\t\\tbreak\\\")\"", "r");
-  fgets(python_path, PATH_MAX, fp);
+  (void)fgets(python_path, PATH_MAX, fp);
   pclose(fp);
 
   if (strlen(python_path) == 0) {


FrancescAlted:
Indeed.  We should ensure that the return value of `fgets` is not NULL.  Will do in another commit.

DimitriPapadopoulos:
Do we really care about the return value of `fgets`?

The code currently tests whether `python_path` has been modified in a useful way:
```c
  if (strlen(python_path) == 0) {
    BLOSC_TRACE_ERROR("Could not find python path");
```
I would expect the value of `strlen(python_path)` to remain `0` in case of error, as initialised here, although nothing in the standard guarantees that:
```c
char python_path[PATH_MAX] = {0};
```
Any way, in the end we're only interested in:
```c
  return dlopen(path, RTLD_LAZY);
```

FrancescAlted:
Yes, I think you are right.  Merging.

DimitriPapadopoulos:
Perhaps an error message when `fgets` fails might give valuable information on the causes of the more general failure to retrieve a path - although I don't expect that to happen often. Perhaps it's worth emitting a specific error message - unlike the case of a `pclose` failure for example.

But that would be for another PR.

DimitriPapadopoulos:
The warning is still there. Strange, I had tested it and was certain everything's OK, but that's not the case:
```c
/my/path/c-blosc2/include/blosc2/plugins-utils.h:44:9: warning: ignoring return value of ‘fgets’ declared with attribute ‘warn_unused_result’ [-Wunused-result]
   44 |   (void)fgets(python_path, PATH_MAX, fp);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
I think we will really have to check the return value...

PR #496, opened 2023-05-15
Stop supporting ancient versions of Visual Studio

### LINKED ISSUE #493: `_MSC_VER` and associated comments: need to support ancient Visual Studio versions?
**Describe the bug**

According to the [Microsoft Visual C++](https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B) article in Wikipedia, the value of `_MSC_VER` is 1600 for Visual compiler **10.0**, not 16.0:
https://github.com/Blosc/c-blosc2/blob/c51d050dfa154411d776d84771fd74ca83bd232b/blosc/schunk.c#L30-L35

From the table there:

| product name                       | product Version | marketing year | `_MSC_VER` |
|------------------------------------|-----------------|----------------|----------|
|                                    |                 |                |          |
| Visual Studio 2008 SP1 [9.0]       | 9.0             | 2010           | 1500     |
| Visual Studio 2010 Beta 1 [10.0]   | 10.0            | 2010           | 1600     |
| visual Studio 2010 Beta 2 [10.0]   | 10.0            | 2010           | 1600     |
| Visual Studio 2010 [10.0]          | 10.0            | 2010           | 1600     |
| Visual Studio 2010 SP1 [10.0]      | 10.0            | 2010           | 1600     |
| Visual Studio 2012 [11.0]          | 11.0            | 2012           | 1700     |
|                                    |                 |                |          |
| Visual Studio 2017 version 15.9.11 | 15.9.11         | 2017           | 1916     |
| Visual Studio 2019 version 16.0.0  | 16.0.0          | 2019           | 1920     |
|                                    |                 |                |          |
| Visual Studio 2019 version 16.1.2  | 16.1.2          | 2019           | 1921     |
|                                    |                 |                |          |

**Expected behavior**

I see no need to support VisualStudio 2010 in a recent project such as _c-blosc2_.

If the need exists, at least fix the comments and change `VS2010 (VC++ 16.0)` to `VS 2010 (VC++ 10.0)`.

**System information:**
 - OS: [e.g. OSX]
 - Compiler [e.g. gcc, clang]
 - Version [e.g. 2.0.1]

**Additional context**

DimitriPapadopoulos:
This removes support for `_MSC_VER <= 1600`, that is Visual Studio 2010 (version 10.0) or earlier.

We expect ends users will run Visual Studio 2012 (version 11.0) or later, which means `_MSC_VER >= 1700`. This is not to say Studio 2012 (version 11.0) will work our of the box, for lack of testing, just that anything earlier will not work.

Fixes #493.


DimitriPapadopoulos:
Should I add something like this in `blosc2.h`?
```c
#if defined(_MSC_VER) && (_MSC_VER < 1700)
#error "Unsupported version of Visual Studio, use Visual Studio 2012 or later."
#endif
```

FrancescAlted:
> Should I add something like this in `blosc2.h`?
> 
> ```c
> #if defined(_MSC_VER) && (_MSC_VER < 1700)
> #error "Unsupported version of Visual Studio, use Visual Studio 2012 or later."
> #endif
> ```

I'm not a windows users, so I don't know, really.  Which is your opinion on a decently modern VS version?

FrancescAlted:
> > Should I add something like this in `blosc2.h`?
> > ```c
> > #if defined(_MSC_VER) && (_MSC_VER < 1700)
> > #error "Unsupported version of Visual Studio, use Visual Studio 2012 or later."
> > #endif
> > ```
> 
> I'm not a windows users, so I don't know, really. Which is your opinion on a decently modern VS version?

Uh, sorry, I did not understand the question.  So, if what you are proposing is supporting at least Visual Studio 2012, I think is a very safe assumption, so +1.

BTW, if we get rid of `win32/stdint-windows.h`, then we can remove https://github.com/Blosc/c-blosc2/blob/main/LICENSES/STDINT.txt too.

DimitriPapadopoulos:
Will remove `LICENSES/STDINT.txt` soon.

FrancescAlted:
LGTM.  Thanks @DimitriPapadopoulos !

Issue #500, opened 2023-05-18
inline function `swap_store` in `blosc2.h`


DimitriPapadopoulos:
**Describe the bug**

Why is the inline function `swap_store` part of the API?

It is mostly used for `int32_t` and `int64_t`. In that case, calls to `malloc` and `memcpy` are overkill. Just use:
```c
uint32_t swap32(uint32_t x) {
  return ((x & 0x000000FF) << 24) | ((x & 0x0000FF00) <<  8) |
         ((x & 0x00FF0000) >>  8) | ((x & 0xFF000000) >> 24);
}

uint64_t _bswap64(uint64_t x) {
  return ((x & 0x00000000000000FF) << 56) | ((x & 0x000000000000FF00) << 40) | ((x & 0x0000000000FF0000) << 24) | ((x & 0x00000000FF000000) <<  8) | 
         ((x & 0x000000FF00000000) >>  8) | ((x & 0x0000FF0000000000) >> 24) | ((x & 0x00FF000000000000) >> 40) | ((x & 0xFF00000000000000) >> 56);
}
```

Actually, you should use compiler intrinsics if they exist:
|       | 2 bytes                                                                                                                              | 4 bytes                                                                                                                             | 8 bytes                                                                                                                              |
|-------|--------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------|
| MSVC  | [_byteswap_ushort](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/byteswap-uint64-byteswap-ulong-byteswap-ushort) | [_byteswap_ulong](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/byteswap-uint64-byteswap-ulong-byteswap-ushort) | [_byteswap_uint64](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/byteswap-uint64-byteswap-ulong-byteswap-ushort) |
[...]

PR #504, opened 2023-05-18
Clean up use of 'extern "C"' in header files

DimitriPapadopoulos:
Private header files are only seen by the C compiler that builds the library. They don't need `extern "C"`.

Public header files can be seen by a C++ compiler. They do need the `extern "C"`. And it's best done individually in each public header file. On the other hand, header files that just use directives of the preprocessor need not `extern "C"`.

### DIFF
--- a/blosc/b2nd_utils.h
+++ b/blosc/b2nd_utils.h
@@ -14,10 +14,6 @@
 #include <b2nd.h>
 #include <../plugins/plugin_utils.h>
 
-#ifdef __cplusplus
-extern "C" {
-#endif
-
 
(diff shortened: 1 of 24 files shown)


DimitriPapadopoulos:
I can also add these to all header files. However, that's useless for private files, since these files are supposed to be compiled only by the C compiler that builds c-blosc2.

I am under the impression that the API of c-blosc2 is not well-delimited, and that you may have external code, or at least plugins, include private header files (https://github.com/Blosc/c-blosc2/issues/500#issuecomment-1553078718). In that case, it might make sent to protect private header files too, in case you include them from C++ code.

PR #505, opened 2023-05-18
Forgot to pclose() in all cases in ca60217

DimitriPapadopoulos:
Error handling would be easier and more readable with `goto`. It helps release resources in the right order at the end of the function with a single return point. The alternative would be a bunch of nested `if`'s which impair readability and result in very long liens because of indention.

See for example:
* [MEM12-C. Consider using a goto chain when leaving a function on error when using and releasing resources](https://wiki.sei.cmu.edu/confluence/display/c/MEM12-C.+Consider+using+a+goto+chain+when+leaving+a+function+on+error+when+using+and+releasing+resources)
* https://lkml.org/lkml/2003/1/12/203


FrancescAlted:
Well seen.  Thanks!

PR #507, opened 2023-05-23
Fix incorrect nested ifdef's from a3e56ee

DimitriPapadopoulos:
Sorry about that. I was focused on keeping a minimal number of `#if`'s in a3e56ee, but I need to add a second `#if defined(_WIN32)` for both correctness and readability.


FrancescAlted:
LGTM.  Thanks!

PR #508, opened 2023-05-23
Clean up header file inclusion, sequel of c4f28da

DimitriPapadopoulos:
* Use "top-down" header file inclusion ordering
* Add missing header files
* Remove useless header files
    
It should be possible to include header files in any order, but we're probably not there yet.

Also, remove a `BLOSC_EXPORT` from a source file, it should be used in the header file when declaring the function.


plugins/codecs/codecs-registry.c
```diff
@@ -4,46 +4,47 @@
...
 void register_codecs(void) {
 
   blosc2_codec ndlz;
   ndlz.compcode = BLOSC_CODEC_NDLZ;
   ndlz.version = 1;
   ndlz.complib = BLOSC_CODEC_NDLZ;
-  ndlz.encoder = ndlz_compress;
-  ndlz.decoder = ndlz_decompress;
+  ndlz.encoder = &ndlz_compress;
+  ndlz.decoder = &ndlz_decompress;
```
FrancescAlted, line 20: This is a serious fix.  I wonder why it was working before (?).

FrancescAlted (commented):
This is a great PR.  I just have put a question, in case you can shed some light on it.

plugins/codecs/codecs-registry.c
DimitriPapadopoulos: Actually, they are equivalent. A function does not have a “value” so C compilers always interpret that kind of expressions as a pointer to the function. From the C standard:
> A _function designator_ is an expression that has function type. Except when it is the operand of the `sizeof` operator, the `_Alignof` operator, or the unary `&` operator, a function designator with _type_ ‘‘function returning type’’ is converted to an expression that has type ‘‘pointer to function returning _type_’’.

I think most C style manuals recommend the `&` to stress we store a pointer to the function, but it is not compulsory. I seem to recall the '&' is usually dropped in C++.

FrancescAlted: Cool!  Makes sense now.

FrancescAlted (approved):

FrancescAlted:
Thanks @DimitriPapadopoulos !

PR #510, opened 2023-05-23
Fix compiler warning

DimitriPapadopoulos:
Typical message:
```
examples/frame_roundtrip.c:58:41: warning: format ‘%lld’ expects argument of type ‘long long int’, but argument 2 has type ‘int64_t’ {aka ‘long int’} [-Wformat=]
   58 |   printf("nbytes, cbytes for schunk: %lld, %lld \n", schunk->nbytes, schunk->cbytes);
      |                                      ~~~^            ~~~~~~~~~~~~~~
      |                                         |                  |
      |                                         long long int      int64_t {aka long int}
      |                                      %ld
```

### DIFF
--- a/examples/frame_roundtrip.c
+++ b/examples/frame_roundtrip.c
@@ -55,7 +55,8 @@ int main(void) {
       goto failed;
     }
   }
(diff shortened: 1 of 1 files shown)


FrancescAlted:
LGTM.  Thanks.

PR #512, opened 2023-05-23
Revert previous changes to vendored code

DimitriPapadopoulos:


FrancescAlted:
Uh, the bitshuffle codec here is actually a [copy of the upstream](https://github.com/kiyo-masui/bitshuffle/blob/b9a1546133959298c56eee686932dbb18ff80f7a/src/bitshuffle_core.c#L103).  Frankly, for future updates in upstream, I'd prefer keeping this code in sync.

FrancescAlted:
> Indeed, I will revert all previous changes. That said, perhaps the block below should be kept in this file, in addition to the Blosc block:
> 
> ```
> /*
>  * Bitshuffle - Filter for improving compression of typed binary data.
>  *
>  * Author: Kiyoshi Masui <kiyo@physics.ubc.ca>
>  * Website: http://www.github.com/kiyo-masui/bitshuffle
>  * Created: 2014
>  *
>  * See LICENSE file for details about copyright and rights to use.
>  *
>  */
> ```
> 
> The rationale is both to warn that the code is vendored, and to comply with the licence:
> 
> > The above copyright notice and this permission notice shall be included in
> > all copies or substantial portions of the Software.

Agreed!

FrancescAlted:
> Most of 3rd party code lives under [`internal-complibs`](https://github.com/Blosc/c-blosc2/tree/main/internal-complibs).
> 
> Are these 3 macros the only piece of code borrowed from [kiyo-masui/bitshuffle](https://github.com/kiyo-masui/bitshuffle)? Is there vendored code elsewhere too?

No, all the bitshuffle code is from the bitshuffle project (actually, Kiyo helped us in integrating it better in C-Blosc).

DimitriPapadopoulos:
It would be clearer if files could be just copied over from [kiyo-masui/bitshuffle](https://github.com/kiyo-masui/bitshuffle), but https://github.com/kiyo-masui/bitshuffle/issues/11 seems to imply it's not straightforward.

DimitriPapadopoulos:
Should I revert changes to other files? In most `bitshuffle-*.[hc]` files, I changed header inclusion stuff, but in a few files, I modified macros.

PR #515, opened 2023-05-27
Pedantic suggestions by checkpatch.pl

DimitriPapadopoulos:
Some suggestions by [`checkpatch.pl`](https://github.com/torvalds/linux/blob/master/scripts/checkpatch.pl), I picked those that don't modify formatting much.


FrancescAlted:
LGTM.  Thanks @DimitriPapadopoulos !

PR #517, opened 2023-06-15
Use zlib CMake target if it already exists

pablode:
If zlib or zlib-ng have already been included in the project using an `add_subdirectory` call, use the provided target and don't try to find the library again.


pablode:
Ah, I just realized that the internal fallback includes zlib-ng as a subdirectory too, which causes the conflicts. Besides, it seems like `HAVE_ZLIB_NG` would need to be passed from the outside, since it can't be figured out from the zlib target alone. Currently do not have the capacity to investigate this further -- closing for now.

PR #523, opened 2023-06-27
Assorted minor documentation fixes

ivilata:
These are very minor fixes from things that I spotted while reading the Manual.


FrancescAlted (approved):
That looks great.  Thanks @ivilata !

PR #525, opened 2023-06-30
CMake: Quote $ENV{CMAKE_OSX_ARCHITECTURES}

ax3l:
For some reason, we see build issues in cibuildwheel with this when it is not quoted. Probably when defines as an empty string it throws:
```
...
  -- Using ZSTD internal sources.
      "STREQUAL" "arm64"
  -- Configuring incomplete, errors occurred!

    Unknown arguments specified
```


ax3l:
X-ref a8ded8512c235758cdfc707d9e4e1d65f3911955

cc @FrancescAlted :)

FrancescAlted (approved):
LGTM

Issue #526, opened 2023-06-30
C-Blosc 2.9.0 w/ MSVC: use of designated initializers requires at least '/std:c++20'


ax3l:
**Describe the bug**
Compiling the downstream project ADIOS2 v2.9.0 with C-Blosc 2.9.0 throws the error:
```
  C:\Program Files (x86)\blosc\include\blosc2.h(1056,5): error C7555: use of designated initializers requires at least '/std:c++20'
```

**To Reproduce**
Compile [ADIOS2](https://github.com/ornladios/ADIOS2) v2.9.0 with `-DADIOS2_USE_Blosc2=ON` on Windows.

E.g.
```cmd
  curl -sLo blosc2-2.9.3.zip ^
    https://github.com/Blosc/c-blosc2/archive/refs/tags/v2.9.3.zip
  powershell Expand-Archive blosc2-2.9.3.zip -DestinationPath dep-blosc2

  :: https://github.com/Blosc/c-blosc2/pull/525
  curl -sLo c-blosc2-cmake.patch ^
    https://patch-diff.githubusercontent.com/raw/Blosc/c-blosc2/pull/525.patch
  python -m patch -p 1 -d dep-blosc2/c-blosc2-2.9.3 c-blosc2-cmake.patch

  cmake -S dep-blosc2/c-blosc2-2.9.3 -B build-blosc2 ^
    -DCMAKE_BUILD_TYPE=Release  ^
    -DBUILD_SHARED=OFF          ^
    -DBUILD_STATIC=ON           ^
    -DBUILD_BENCHMARKS=OFF      ^
    -DBUILD_EXAMPLES=OFF        ^
    -DBUILD_FUZZERS=OFF         ^
    -DBUILD_TESTS=OFF           ^
    -DPREFER_EXTERNAL_ZLIB=ON   ^
    -DZLIB_USE_STATIC_LIBS=ON

  curl -sLo adios2-2.9.0.zip ^
    https://github.com/ornladios/ADIOS2/archive/v2.9.0.zip
  powershell Expand-Archive adios2-2.9.0.zip -DestinationPath dep-adios2

  cmake -S dep-adios2/ADIOS2-2.9.0 -B build-adios2 ^
    -DCMAKE_BUILD_TYPE=Release  ^
    -DBUILD_SHARED_LIBS=OFF     ^
    -DBUILD_TESTING=OFF         ^
    -DADIOS2_BUILD_EXAMPLES=OFF ^
[...]

PR #527, opened 2023-06-30
blosc2.h: Avoid C++20 Designated Init

ax3l:
See #526.
Designated initializers are in C99 but only in C++20, too.


FrancescAlted (approved):
LGTM.  Thanks @ax3l !

PR #529, opened 2023-07-01
Fix: CMake External Zlib

ax3l:
I see downstream linkage issues with externally built, static zlib library on Windows. This tries to fix it.

Docs:
- https://cmake.org/cmake/help/v3.27/module/FindZLIB.html

Generally, we should use `ZLIB::ZLIB` as the target to link instead of decluttering these result variables.

- [ ] testing in https://github.com/openPMD/openPMD-api/pull/1387


ax3l:
Although these two lines area clearly mistyped compared to the `FindZlib.cmake` docs, this does not yet solve my problem. Maybe the `zlib.lib` is not propagated to dependents...?

FrancescAlted (approved):
LGTM.  Thanks @ax3l !

FrancescAlted:
Thanks for the fix.  Tell me if you need something more to fix your issue.  For the time being, I am going to merge this one.

Issue #528, opened 2023-07-01
Incorrect roundtrip for `[blosc2.Filter.BITSHUFFLE, blosc2.Filter.BYTEDELTA]`


froody:
**Describe the bug**
When invoked from python with a random uint16 array with max value of 118, the roundtrip for bitshuffle + bytedelta is incorrect, even though they are correct on their own.

**To Reproduce**
```
import numpy as np
import blosc2
np.random.seed(0)

a = np.random.randint(118, size=13058819, dtype=np.uint16)

def test_filters(a, filters):
    c = blosc2.pack_array2(a, cparams={"filters":filters})
    u = blosc2.unpack_array2(c)
    if not np.array_equal(a, u):
        print("Error: arrays are not equal with", filters)

test_filters(a, [blosc2.Filter.BITSHUFFLE, blosc2.Filter.BYTEDELTA])
test_filters(a, [blosc2.Filter.BITSHUFFLE])
test_filters(a, [blosc2.Filter.BYTEDELTA])
test_filters(a, [blosc2.Filter.SHUFFLE, blosc2.Filter.BYTEDELTA])
```

**Expected behavior**
(nothing printed)

**Logs**
`Error: arrays are not equal with [<Filter.BITSHUFFLE: 2>, <Filter.BYTEDELTA: 34>]`

**System information:**
 - OS: [Ubuntu 22.04
 - Compiler: n/a
 - Version: 2.2.4

**Additional context**

froody:
Repro in C:
* Run with no args to exhibit failure
* Run with any arg to show successfull roundtrip with SHUFFLE + BYTEDELTA
```
#include <stdio.h>
#include <blosc2.h>
#include "blosc2/filters-registry.h"

#define NCHUNKS 1
#define NTHREADS 2
#define TYPESIZE 2
#define LEN 13058819
#define CHUNKSIZE (TYPESIZE * LEN)

int main(int argc, char **argv) {
  blosc2_init();
  uint16_t *ref_data = (uint16_t *)malloc(CHUNKSIZE);
  std::srand(0);
  for (int i = 0; i < LEN; i++) {
    ref_data[i] = std::rand() / (RAND_MAX / 118);
  }
  uint16_t *data_dest = (uint16_t *)malloc(CHUNKSIZE);
  int32_t isize = CHUNKSIZE;
  blosc2_cparams cparams = BLOSC2_CPARAMS_DEFAULTS;
  cparams.compcode = BLOSC_ZSTD;
[...]

PR #530, opened 2023-07-01
Cycle buffers

froody:
Fix for https://github.com/Blosc/c-blosc2/issues/528

I'm unsure about the correctness of this change but I ran all the unit tests and they all passed. The attached test fails before the addition of `_cycle_buffers`  and passes after adding `_cycle_buffers`. 

My rationale for the fix is that all the sites that were swapping buffers looked like they intended to rotate 3 variables, but actually lose a variable each time:
```
_src = _dest;
_dest = _tmp;
_tmp = _src;
```
is equivalent to 
```
uint8_t *scratch = _dest;
_dest = _tmp;
_src = scratch;
_tmp = scratch;
``` 
So that any filter which writes to the `_tmp` buffer will be corrupting its own input.

### DIFF
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -907,6 +907,12 @@ static int blosc2_intialize_header_from_context(blosc2_context* context, blosc_h
   return 0;
 }
 
+void _cycle_buffers(void **src, void **dest, void **tmp) {
+  void *tmp2 = *src;
+  *src = *dest;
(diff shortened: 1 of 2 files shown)


FrancescAlted:
Well seen!  This produces a warning on the params of the new `_cycle_buffers`, but I will fix that later on.  Thanks @froody !

PR #531, opened 2023-07-01
Fix discrepancy between SIMD/Non-SIMD versions of bytedelta

### LINKED ISSUE #524: BYTEDELTA,CPU_HAS_SIMD=1 incompatible with BYTEDELTA,CPU_HAS_SIMD=0
**Describe the bug**
Both the forward and backward pass of bytedelta initialize a `uint8_t _v2 = 0` which has a subtle dual use. Chunks where `stream_len % 16 != 0` result in different output for CPU_HAS_SIMD=0 vs CPU_HAS_SIMD=1.

In the CPU_HAS_SIMD=1 case, `bytes16 v2` is continually updated with the current value of v2, but once the common tail/leftover path is entered, `_v2` is re-initialized to `0` partway through the stream. In the CPU_HAS_SIMD=0 case, `_v2` is continually updated with the rolling value.

The workaround to read data generated by `CPU_HAS_SIMD=1` on systems with `CPU_HAS_SIMD=0` is to reset _v2 to 0 when `stream_len % 16 != 0 && ip % 16 == 0 && ip + 15 > stream_len`, but the ideal fix is to initialize _v2 with the appropriate value from the ﻿[bytes16 v2](https://github.com/Blosc/c-blosc2/blob/main/plugins/filters/bytedelta/bytedelta.c#L98) when CPU_HAS_SIMD=1. Fixing it the ideal way would be backward incompatible with any previously serialized data so it seems like there's no perfect solution.

**To Reproduce**
1. Compile the program below and statically link it to libblosc2, ensure that `CPU_HAS_SIMD == 1` in `plugins/filters/bytedelta/bytedelta.c`. Rename this binary to `bug_simd`.
2. Hardcode CPU_HAS_SIMD = 0 in `filters/bytedelta/bytedelta.c`, recompile libblosc2, and recompile and statically relink the program below. Rename this to `bug_generic`.
3. `./bug_simd 1` # Successful roundtrip data <-> schunk !
4. `./bug_generic 1` # Successful roundtrip data <-> schunk !
5. `./bug_simd` # Decompressed data differs from original 116, 231382041, 231371503!
6. `./bug_simd 1` # Successful roundtrip data <-> schunk !
7. `./bug_generic` # Decompressed data differs from original 116, 231382041, 231392835!
```
#include <stdio.h>
#include <blosc2.h>
#include "blosc2/filters-registry.h"

#define NCHUNKS 1
#define NTHREADS 2
#define TYPESIZE 4
#define CHUNKSIZE (TYPESIZE * 29) // (CHUNKSIZE / TYPESIZE) % 16 must != 0 to exploit this bug.

int32_t ref_data[CHUNKSIZE] = {
  546303464, 823102828, 763542202,  46073357, 639725367, 831718428,
  754955461, 313104751, 556923531, 902333302, 874693449, 244457120,
  42880164, 677450262, 909282669, 716283249, 231382041, 835942381,
  341227635, 956334593, 656239542, 588781027, 805567515, 504325858,
  734152813, 220178903, 436001506, 691382423, 534977848
};

int main(int argc, char **argv) {
  blosc2_init();
  static int32_t data_dest[CHUNKSIZE];
  int32_t isize = CHUNKSIZE * sizeof(int32_t);
  blosc2_cparams cparams = BLOSC2_CPARAMS_DEFAULTS;
  cparams.compcode = BLOSC_ZSTD;
  cparams.filters[BLOSC2_MAX_FILTERS - 1] = BLOSC_FILTER_BYTEDELTA;
  cparams.filters_meta[BLOSC2_MAX_FILTERS - 1] = 7;

  blosc2_dparams dparams = BLOSC2_DPARAMS_DEFAULTS;
  blosc2_schunk* schunk;

  /* Create a super-chunk container */
  cparams.typesize = TYPESIZE;
  cparams.nthreads = NTHREADS;
  dparams.nthreads = NTHREADS;
  char *urlpath = (char *)"/tmp/dir1.b2frame";
  blosc2_storage storage = {.contiguous=false, .urlpath=urlpath, .cparams=&cparams, .dparams=&dparams};

  if (argc > 1) {
    /* Remove the directory and write new data */
    blosc2_remove_dir(storage.urlpath);
    schunk = blosc2_schunk_new(&storage);
    blosc2_schunk_append_buffer(schunk, ref_data, isize);
  }

  /* Read the chunks that were written before */
  schunk = blosc2_schunk_open(urlpath);
  blosc2_schunk_decompress_chunk(schunk, 0, data_dest, isize);
  for (int i = 0; i < CHUNKSIZE; i++) {
    if (data_dest[i] != ref_data[i]) {
      printf("Decompressed data differs from original %d, %d, %d!\n",
             CHUNKSIZE, ref_data[i], data_dest[i]);
      return -1;
    }
  }

  printf("Successful roundtrip data <-> schunk !\n");
  blosc2_schunk_free(schunk);
  blosc2_destroy();
  return 0;
}
```

**Expected behavior**
Steps 3 through 4 each print `Successful roundtrip data <-> schunk !`

**Logs**
N/A

**System information:**
 - OS: Ubunt 22.04
 - Compiler: clang-13
 - Version 0787e1b26027b582aaf8955dc18534689ef4f21b

**Additional context**
N/A

### LINKED ISSUE #528: Incorrect roundtrip for `[blosc2.Filter.BITSHUFFLE, blosc2.Filter.BYTEDELTA]`

froody:
Fixes #524 , but breaks backward compatibility with existing datasets generated by SIMD version. Maybe keep two versions and have a id in the registry? BLOSC_FILTER_BYTEDELTA = 35; BLOSC_FILTER_BYTEDELTA_LEGACY = 34?

### DIFF
--- a/plugins/filters/bytedelta/bytedelta.c
+++ b/plugins/filters/bytedelta/bytedelta.c
@@ -46,6 +46,8 @@ bytes16 simd_prefix_sum(bytes16 x)
   return x;
 }
 
+uint8_t simd_get_last(bytes16 x) { return (_mm_extract_epi16(x, 7) >> 8) & 0xFF; }
+
 #elif defined(__aarch64__) || defined(_M_ARM64)
 // ARM v8 NEON code path
 #define CPU_HAS_SIMD 1
@@ -72,6 +74,8 @@ bytes16 simd_prefix_sum(bytes16 x)
   return x;
 }
 
+uint8_t simd_get_last(bytes16 x) { return vgetq_lane_u8(x, 15); }
+
 #endif
 
 
@@ -93,6 +97,7 @@ int bytedelta_forward(const uint8_t *input, uint8_t *output, int32_t length, uin
   const int stream_len = length / typesize;
   for (int ich = 0; ich < typesize; ++ich) {
     int ip = 0;
+    uint8_t _v2 = 0;
     // SIMD delta within each channel, store
 #if defined(CPU_HAS_SIMD)
     bytes16 v2 = {0};
@@ -104,9 +109,12 @@ int bytedelta_forward(const uint8_t *input, uint8_t *output, int32_t length, uin
       output += 16;
       v2 = v;
     }
+    if (stream_len > 15) {
+      _v2 = simd_get_last(v2);
+
+    }
 #endif // #if defined(CPU_HAS_SIMD)
     // scalar leftover
-    uint8_t _v2 = 0;
     for (; ip < stream_len ; ip++) {
       uint8_t v = *input;
       input++;
@@ -137,6 +145,7 @@ int bytedelta_backward(const uint8_t *input, uint8_t *output, int32_t length, ui
   const int stream_len = length / typesize;
   for (int ich = 0; ich < typesize; ++ich) {
     int ip = 0;
+    uint8_t _v2 = 0;
     // SIMD fetch 16 bytes from each channel, prefix-sum un-delta
 #if defined(CPU_HAS_SIMD)
     bytes16 v2 = {0};
@@ -148,9 +157,11 @@ int bytedelta_backward(const uint8_t *input, uint8_t *output, int32_t length, ui
       simd_store(output, v2);
       output += 16;
     }
+    if (stream_len > 15) {
+      _v2 = simd_get_last(v2);
+    }
 #endif // #if defined(CPU_HAS_SIMD)
     // scalar leftover
-    uint8_t _v2 = 0;
     for (; ip < stream_len; ip++) {
       uint8_t v = *input + _v2;
       input++;
--- a/plugins/filters/bytedelta/test_bytedelta.c
+++ b/plugins/filters/bytedelta/test_bytedelta.c
@@ -15,17 +15,29 @@
     To run:
 
     $ ./test_bytedelta
+    Testing bytedelta with write_correct=1, read_correct=1
+    Testing bytedelta with write_correct=0, read_correct=1
+    Testing bytedelta with write_correct=1, read_correct=0
+    Testing bytedelta with write_correct=0, read_correct=0
     Successful roundtrip!
-    Compression: 41472 -> 4937 (8.4x)
-    rand: 36535 obtained
+    Compression: 41472 -> 5599 (7.4x)
+    rand: saved 35873 bytes
 
+    Testing bytedelta with write_correct=1, read_correct=1
+    Testing bytedelta with write_correct=0, read_correct=1
+    Testing bytedelta with write_correct=1, read_correct=0
+    Testing bytedelta with write_correct=0, read_correct=0
     Successful roundtrip!
-    Compression: 1792 -> 1005 (1.8x)
-    mixed_values: 787 obtained
+    Compression: 1792 -> 499 (3.6x)
+    mixed_values: saved 1293 bytes
 
+    Testing bytedelta with write_correct=1, read_correct=1
+    Testing bytedelta with write_correct=0, read_correct=1
+    Testing bytedelta with write_correct=1, read_correct=0
+    Testing bytedelta with write_correct=0, read_correct=0
     Successful roundtrip!
-    Compression: 16128 -> 1599 (10.1x)
-    arange_like: 14529 obtained
+    Compression: 16128 -> 1473 (10.9x)
+    arange_like: saved 14655 bytes
 
 **********************************************************************/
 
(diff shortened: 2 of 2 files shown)


FrancescAlted:
This looks good to me.  Could you add a test that fails with the previous version of the filter?

Also, for guaranteeing backward compatibility, we can play with the [filter_meta](https://github.com/Blosc/c-blosc2/blob/bcc3825d2203a5f07af9590a7bec178b06e05778/README_CHUNK_FORMAT.rst#header) in the header.  By default, a [0 is stored there](https://github.com/Blosc/c-blosc2/blob/bcc3825d2203a5f07af9590a7bec178b06e05778/blosc/blosc2.c#L871), but starting from this fix, we can set the `filter_meta` for bytedelta to 1.  Then, code can be added to check for the filter_meta for bytedelta and deal with different scenarios (0 means buggy version, 1 means fixed version).  How that sounds?

FrancescAlted:
Also, it looks like there is a typo in https://github.com/Blosc/c-blosc2/pull/531/files#diff-5dff8fdab0f3f4c0bd14551691ffe87dba0d6e5ecd780041c3f94af42714dfd2R77

froody:
> but starting from this fix, we can set the `filter_meta` for bytedelta to 1

How exactly do I do that without a broader refactor? As far as I can tell I would need to put a special check in the body of `blosc_compress_context` which feels hacky to me.

FrancescAlted:
> > but starting from this fix, we can set the `filter_meta` for bytedelta to 1
> 
> How exactly do I do that without a broader refactor? As far as I can tell I would need to put a special case in the body of `blosc_compress_context` to set `filters_meta[i] = 1` where `filters[i]= BLOSC_FILTER_BYTEDELTA which feels hacky to me.
> 
> There's also the problem that a nonzero value of `meta` in `bytedelta_forward` and `bytedelta_backward` overrides the bytesize. What about existing data with a serialized `meta` value of 1? That would lead to corrupt decompression after applying these changes.

After some consideration, I think you are right.  Let's forget about using `meta` for this.

> 
> I would love to have access to the `version` field of `blosc_header` but that's not possible without a refactor.

It is actually possible.  Version is the first byte written to the blosc header, so with `input[0]` you can have access to this.  Only works for backward operation, but this is what you want.

> 
> Either way I recommend to increment the version in the header [here](https://github.com/Blosc/c-blosc2/blob/bcc3825d2203a5f07af9590a7bec178b06e05778/include/blosc2.h#L126-L139) to prevent old versions of the library from decompressing chunks written by newer versions.

I concur, and I am in favor of this.  Would you like to increment the current version to 6 and call the symbol for the new version something like `BLOSC2_VERSION_FORMAT_STABLE_POST1 ` and then make `BLOSC2_VERSION_FORMAT` an alias of this one (instead of current `BLOSC2_VERSION_FORMAT_STABLE `) ?  Then, by using the new version in the backward operation, we can easily distinguish between one format or another. Bonus points is you can provide a file compressed with the 'bad' filter version and check that newer versions of Blosc2 can still read it. 

We have considered all the options (including introducing backward incompatibility and having another, corrected, filter ID), but increasing the version of the header seems the simplest to us.

Thanks in advance!

FrancescAlted:
Oops, I was wrong, and the backward filter does not currently have access to the header version, so using that is not going to work.

In #532 I have explored adding a new filter ID, which seems to me the least resistance path for a fix.  However, as it stands now, #532 still requires some additional tweaking.

PR #536, opened 2023-07-25
CMake: Static lib Deps are Public

ax3l:
In `target_link_libraries`, we need to link dependencies as `PUBLIC` to make sure they are propagated downstream.

Seen when building a static c-blosc2 library, depending on an external, static zlib library and using downstream in ADIOS2.

Follow-up to #529

### DIFF
--- a/blosc/CMakeLists.txt
+++ b/blosc/CMakeLists.txt
@@ -304,7 +304,7 @@ if(BUILD_STATIC)
     if(MSVC OR MINGW)
         set_target_properties(blosc2_static PROPERTIES PREFIX lib)
     endif()
-    target_link_libraries(blosc2_static ${LIBS})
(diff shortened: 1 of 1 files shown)


ax3l:
The `pkgconfig/blosc2.pc` file should also contain the static dependency libs for the shared libs (or we need a 2nd pc file, becase c-blosc2 supports building two PC files at once).

@FrancescAlted is there a reason c-blosc does not yet add a `Blosc2Config.cmake` file when installed? :)

ax3l:
> @FrancescAlted is there a reason c-blosc does not yet add a Blosc2Config.cmake file when installed? :)

Ok, starting one now in #537

ax3l:
This PR on it's own is not sufficient and #537 solves the problem holistically :)

PR #537, opened 2023-07-25
CMake: Install Blosc2Config.cmake 

ax3l:
This adds a [CMake installer](https://cliutils.gitlab.io/modern-cmake/chapters/install/installing.html) to conserve targets and properties on install, so CMake users do not need to write `FindBlosc2.cmake` files anymore.

```bash
export CMAKE_PREFIX_PATH=$HOME/sw/c-blosc2-installed-here
```
```cmake
# ...
find_package(Blosc2 2.4 CONFIG)
```

This also helps to preserve transitive dependencies on CMake targets, especially useful for fully static builds, e.g., for Python wheels.

The installed config package is designed to work both on Unix (Linux/macOS) and Windows.

- [x] testing with downstream projects ADIOS2 -> openPMD-api
  https://github.com/openPMD/openPMD-api/pull/1387


ax3l:
For @FrancescAlted - please let me know if you prefer more inline comments or need more details.

I am also happy to do a few general updates/modernizations to CMake as a follow-up PR. The PR here tries to focus on only changing the things necessary for now, enabling all-static builds and propagating CMake targets and their dependencies to downstream developers cleanly.

FrancescAlted (changes requested):
That looks great to me.  However, as some changes are quite large, I would appreciate some more comments in the places that things have changed most.  Thanks in advance!

FrancescAlted:
> For @FrancescAlted - please let me know if you prefer more inline comments or need more details.

Yes, I have suggested that already :-)

> 
> I am also happy to do a few general updates/modernizations to CMake as a follow-up PR. The PR here tries to focus on only changing the things necessary for now, enabling all-static builds and propagating CMake targets and their dependencies to downstream developers cleanly.

Sounds good to me.  Let's first merge this one and proceed gradually.  Thanks!

ax3l:
> as some changes are quite large, I would appreciate some more comments in the places that things have changed most. Thanks in advance!

Thank you @FrancescAlted! I added more inline comments and added some cleaning in the last commit.
Please feel free to add inline comments for sections that are not yet self-explanatory :)

ax3l:
Attn @FrancescAlted :)

FrancescAlted (approved):
LGTM.  Thanks!

ax3l:
My pleasure! :)
Thank you for merging!

Issue #539, opened 2023-08-01
CMake failure when trying to use system zlib-ng: ZLIB_NG::ZLIB_NG target not found


mgorny:
**Describe the bug**
If `PREFER_EXTERNAL_ZLIB` option is enabled, c-blosc2 attempts to use `zlib-ng`, then plain `zlib`. However, if `find_package()` succeeds for `zlib-ng`, CMake fails because of missing targets:

```
$ cmake .. -G Ninja -DPREFER_EXTERNAL_ZLIB=ON
-- The C compiler identification is GNU 13.2.0
-- The ASM compiler identification is GNU
-- Found assembler: /usr/bin/cc
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
Configuring for Blosc version: 2.10.2.dev
-- Using LZ4 internal sources.
-- Found ZLIB_NG: /usr/lib64/libz-ng.so  
-- Found zlib-ng: /usr/lib64/libz-ng.so, /usr/include
-- Using ZSTD internal sources.
-- No build type specified. Defaulting to 'Release'.
-- Building for system processor x86_64
-- Building for compiler ID GNU
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Adding support for assembly sources in ZSTD
-- Adding run-time support for SSE2
-- Adding run-time support for AVX2
Skipping test_shuffle_roundtrip_altivec on non-ALTIVEC builds
Skipping test_shuffle_roundtrip_neon on non-NEON builds
-- Configuring done (0.2s)
CMake Error at blosc/CMakeLists.txt:108 (target_link_libraries):
  Target "blosc2_shared" links to:

    ZLIB_NG::ZLIB_NG

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.



CMake Error at blosc/CMakeLists.txt:111 (target_link_libraries):
  Target "blosc2_static" links to:

    ZLIB_NG::ZLIB_NG

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.



[...]

PR #541, opened 2023-08-18
CMake: Update FindZLIB_NG For Target

ax3l:
Generate an import target to bring `FindZLIB_NG.cmake` up to par with the CMake-default shipped `FindZLIB.cmake`.

Follow-up to #537 for external ZLIB-NG dependencies.

Seen in https://github.com/conda-forge/c-blosc2-feedstock/pull/42

### DIFF
--- a/cmake/FindZLIB_NG.cmake
+++ b/cmake/FindZLIB_NG.cmake
@@ -27,9 +27,21 @@ else(ZLIB_NG_INCLUDE_DIR AND ZLIB_NG_LIBRARIES)
   set(ZLIB_NG_FOUND FALSE)
 endif(ZLIB_NG_INCLUDE_DIR AND ZLIB_NG_LIBRARIES)
 
+# generate CMake target
 if(ZLIB_NG_FOUND)
   message(STATUS "Found zlib-ng: ${ZLIB_NG_LIBRARIES}, ${ZLIB_NG_INCLUDE_DIR}")
-endif(ZLIB_NG_FOUND)
+
+  set(ZLIB_NG_INCLUDE_DIRS ${ZLIB_NG_INCLUDE_DIR})
+
+  if(NOT TARGET ZLIB_NG::ZLIB_NG)
+    add_library(ZLIB_NG::ZLIB_NG UNKNOWN IMPORTED)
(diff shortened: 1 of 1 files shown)


ax3l:
@FrancescAlted finished now :)
(I pushed a few style updates after opening the PR, to keep the same spacing as already used in the file).

FrancescAlted:
Ok.  The ` ZLIB_NG::ZLIB_NG` syntax is beyond my knowledge.  Could you please explain why do we need it now and not before?  Maybe this ould fix https://github.com/Blosc/c-blosc2/issues/539 too?

ax3l:
Hi @FrancescAlted,

Yes, let me get into this:
In CMake 3+, one tries to avoid passing a lot of arrays as individual variables. Instead, one collects them in "CMake targets" that have properties, which are defined include lists, library names, compile options, etc.

All the `Find*.cmake` scripts in CMake were modernized to provide those targets, which can simply be used via `target_link_libraries(our_target PUBLIC dependency)`. This is essentially the same as C/C++ structs, but with well-defined member names, which avoid the usual issues of naming things (_DIR or _DIRS etc.) and simplifies propagating includes, libs, requirements, defines, etc. in a single call.

The syntax `Dependency::` is added for installed targets, e.g., in `DependencyConfig.cmake` files (and `FindDependency.cmake` files if no such config file exists) to encapsulate them in a "namespace", just like C++ namespaces.

Here are a few more resources if you like to dig deeper:
- https://hsf-training.github.io/hsf-training-cmake-webpage/04-targets/index.html
- https://cliutils.gitlab.io/modern-cmake/chapters/basics.html
- https://cliutils.gitlab.io/modern-cmake/chapters/install.html
- https://cmake.org/cmake/help/latest/command/set_target_properties.html

Let me know if you like have more questions!

FrancescAlted:
@ax3l , many thanks for such a detailed explanation!  I'm not that experienced in cmake science, so it is great that somebody more knowledgeable give us a hand to modernize these parts. (BTW, we already have a book called "Modern CMake for C++ ", and it is time for us to have a serious look into it).  Thanks again, we really appreciate your help.

PR #542, opened 2023-08-18
Fix CMake Install: Find*.cmake Modules

ax3l:
Install the CMake modules in `cmake/` into our install prefix. If we build with external dependencies for LZ4 and ZLIB_NG, then downstream projects will need those Modules to populate our CMake import targets in `Blosc2Config.cmake` during the `find_dependency` calls (which are just wrappers for `find_package`).

Follow-up to #537 and #541.

### DIFF
--- a/Blosc2Config.cmake.in
+++ b/Blosc2Config.cmake.in
@@ -9,7 +9,7 @@ if(POLICY CMP0074)
 endif()
 
 # locate the installed FindABC.cmake modules
-list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake")
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/Modules")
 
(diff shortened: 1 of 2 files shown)


ax3l:
@FrancescAlted this is ready to go and tested with https://github.com/conda-forge/c-blosc2-feedstock/pull/42 and ADIOS2 as downstream dependency that consumes C-Blosc2 + transitive external dependencies.

If you like, maybe we could tag a new patch release and skip 2.10.1 on CF in favor of 2.10.2?
Alternatively, I can show you how to patch these two PRs in to https://github.com/conda-forge/c-blosc2-feedstock/pull/42

ax3l:
P.S.: There is a typo in my last name in https://github.com/Blosc/c-blosc2/releases :)

FrancescAlted:
> P.S.: There is a typo in my last name in https://github.com/Blosc/c-blosc2/releases/tag/v2.10.1 :)

Oops.  Fixed!

PR #543, opened 2023-08-18
Fix redefinition of PATH_MAX macro with mingw

Biswa96:
This fixes the following warning with mingw gcc

    blosc-private.h:187: warning: "PATH_MAX" redefined
      187 | #define PATH_MAX MAX_PATH
          |
    limits.h:20: note: this is the location of the previous definition
       20 | #define PATH_MAX        260
          |


FrancescAlted:
Thanks @Biswa96 !

PR #545, opened 2023-08-19
Fix unused parameter warning with dlopen function

Biswa96:
This fixes the following warning with mingw gcc compiler

    blosc-private.h:201:55: warning: unused parameter 'flags' [-Wunused-parameter]
      201 | static inline void *dlopen (const char *filename, int flags) {
          |                                                   ~~~~^~~~~


FrancescAlted (approved):
Thanks @Biswa96 !

PR #547, opened 2023-08-20
Fetch Content Superbuild for zlib-ng

ax3l:
Add additional logic needed to make superbuilds with fetch-content on recent versions of zlib-ng work in #544.


FrancescAlted:
Hey @ax3l this is really cool!  It is the first time that we are using `FetchContent` machinery in cmake, but I think this looks great.  I'm going to merge.  If that works well, we may want to use the same approach for zstd too.

ax3l:
Hi @FrancescAlted, yes it's fun! :tada:
That sounds great, yes let's do that for all the things we have an internal fallback?

PR #548, opened 2023-08-20
Fix: CMake Config Version w/o Suffix

ax3l:
When writing to CMake, we cannot append the suffix string in the version file, i.e., have to omit `.dev`.

The reason for that is that `VERSION_GREATER_EQUAL` is not that flexible: https://cmake.org/cmake/help/latest/variable/CMAKE_VERSION.html and downstream users comparing against the version will get a `FALSE` result comparing things like
```cmake
if (Blosc2_VERSION VERSION_GREATER_EQUAL 2.10.1)
```
if `Blosc2_VERSION` is set to `2.10.2.dev`.

Follow-up to #537


ax3l:
No, that's wrong. `.dev` suffix is no problem, just tested. No need to change this.

PR #549, opened 2023-08-21
CMake: Cleanup Threads Search & Propagation

ax3l:
Search Threads upfront, because we configure specific options that are cached and reused by dependencies in superbuilds. This brings all `find_package` calls in the same central file before we start including sub-directories with sub-projects.

Set same options as in `find_package` for the `find_dependency` wrapper in `Blosc2Config.cmake`.

This also solves a dependency propagation issue for all-static builds [that I saw on Windows](https://github.com/openPMD/openPMD-api/pull/1515), when Threads are found.


FrancescAlted:
Ok, this is a bit out of reach to my knowledge, but I trust your judgement, so merging.

PR #550, opened 2023-08-21
Always take BLOSC_STRICT_ALIGN path

thesamesam:
Unaligned access is UB, even on x86. UBsan will also detect this (-fsanitize=undefined or -fsanitize=alignment).

On arm, I hit SIGBUSes in pandas's test suite via pandas->pytables->c-blosc2 because of unaligned loads and stores.

Modern compilers are capable of optimising the "slow" path bitshifts and memcpy into faster alternatives where it is legal.

Bug: https://bugs.gentoo.org/911660
Bug: https://github.com/pandas-dev/pandas/issues/54391
Bug: https://github.com/pandas-dev/pandas/issues/54396

### DIFF
--- a/blosc/fastcopy.c
+++ b/blosc/fastcopy.c
@@ -31,6 +31,16 @@
 #define inline __inline  /* Visual C is not C99, but supports some kind of inline */
 #endif
 
+/* Copy LEN bytes (7 or fewer) from FROM into OUT. Return OUT + LEN. */
+static inline unsigned char *copy_bytes(unsigned char *out, const unsigned char *from, unsigned len) {
+  /* We use memcpy nowadays here, but keep the assert given this function should really be
+     replaced with outright memcpy at any instances anyway. */
+  assert(len < 8);
+
+  memcpy(out, from, len);
+
+  return out + len;
+}
 
 static inline unsigned char *copy_1_bytes(unsigned char *out, const unsigned char *from) {
   *out++ = *from;
@@ -38,44 +48,33 @@ static inline unsigned char *copy_1_bytes(unsigned char *out, const unsigned cha
 }
 
 static inline unsigned char *copy_2_bytes(unsigned char *out, const unsigned char *from) {
-  uint16_t chunk;
-  memcpy(&chunk, from, 2);
-  memcpy(out, &chunk, 2);
-  return out + 2;
+  /* These copy_*_bytes helpers are obsolete but we keep them around for compatibility;
+     this should be fine given they're all inline. */
+  return copy_bytes(out, from, 2);
 }
 
 static inline unsigned char *copy_3_bytes(unsigned char *out, const unsigned char *from) {
-  out = copy_1_bytes(out, from);
-  return copy_2_bytes(out, from + 1);
+  return copy_bytes(out, from, 3);
 }
 
 static inline unsigned char *copy_4_bytes(unsigned char *out, const unsigned char *from) {
-  uint32_t chunk;
-  memcpy(&chunk, from, 4);
-  memcpy(out, &chunk, 4);
-  return out + 4;
+  return copy_bytes(out, from, 4);
 }
 
 static inline unsigned char *copy_5_bytes(unsigned char *out, const unsigned char *from) {
-  out = copy_1_bytes(out, from);
-  return copy_4_bytes(out, from + 1);
+  return copy_bytes(out, from, 5);
 }
 
 static inline unsigned char *copy_6_bytes(unsigned char *out, const unsigned char *from) {
-  out = copy_2_bytes(out, from);
-  return copy_4_bytes(out, from + 2);
+  return copy_bytes(out, from, 6);
 }
 
 static inline unsigned char *copy_7_bytes(unsigned char *out, const unsigned char *from) {
-  out = copy_3_bytes(out, from);
-  return copy_4_bytes(out, from + 3);
+  return copy_bytes(out, from, 7);
 }
 
 static inline unsigned char *copy_8_bytes(unsigned char *out, const unsigned char *from) {
-  uint64_t chunk;
-  memcpy(&chunk, from, 8);
-  memcpy(out, &chunk, 8);
-  return out + 8;
+  return copy_bytes(out, from, 8);
 }
 
 
@@ -85,13 +84,10 @@ static inline unsigned char *copy_16_bytes(unsigned char *out, const unsigned ch
   chunk = _mm_loadu_si128((__m128i*)from);
   _mm_storeu_si128((__m128i*)out, chunk);
   out += 16;
-#else
-  int i;
-  for (i = 0; i < 16; i++) {
-    *out++ = *from++;
-  }
-#endif
   return out;
+#else
+  return copy_bytes(out, from, 16);
+#endif
 }
 
 static inline unsigned char *copy_32_bytes(unsigned char *out, const unsigned char *from) {
@@ -100,6 +96,7 @@ static inline unsigned char *copy_32_bytes(unsigned char *out, const unsigned ch
   chunk = _mm256_loadu_si256((__m256i*)from);
   _mm256_storeu_si256((__m256i*)out, chunk);
   out += 32;
+  return out;
 #elif defined(__SSE2__)
   __m128i chunk;
   chunk = _mm_loadu_si128((__m128i*)from);
@@ -108,13 +105,10 @@ static inline unsigned char *copy_32_bytes(unsigned char *out, const unsigned ch
   chunk = _mm_loadu_si128((__m128i*)from);
   _mm_storeu_si128((__m128i*)out, chunk);
   out += 16;
-#else
-  int i;
-  for (i = 0; i < 32; i++) {
-    *out++ = *from++;
-  }
-#endif
   return out;
+#else
+  return copy_bytes(out, from, 32);
+#endif
 }
 
 // This is never used, so comment it out
@@ -126,16 +120,6 @@ static inline unsigned char *copy_32_bytes(unsigned char *out, const unsigned ch
 //  return out + 32;
 //}
 //#endif  // __AVX2__
-
-/* Copy LEN bytes (7 or fewer) from FROM into OUT. Return OUT + LEN. */
-static inline unsigned char *copy_bytes(unsigned char *out, const unsigned char *from, unsigned len) {
-  assert(len < 8);
-
-  while (len--) {
-    *out++ = *from++;
-  }
-  return out;
-}
 
 // Define a symbol for avoiding fall-through warnings emitted by gcc >= 7.0
 #if ((defined(__GNUC__) && BLOSC_GCC_VERSION >= 700) && !defined(__clang__) && \
--- a/blosc/fastcopy.c
+++ b/blosc/fastcopy.c
@@ -22,9 +22,7 @@
 
 #include <assert.h>
 #include <stdint.h>
-#if defined(BLOSC_STRICT_ALIGN)
 #include <string.h>
-#endif
 
 /*
  * Use inlined functions for supported systems.
@@ -40,13 +38,9 @@ static inline unsigned char *copy_1_bytes(unsigned char *out, const unsigned cha
 }
 
 static inline unsigned char *copy_2_bytes(unsigned char *out, const unsigned char *from) {
-#if defined(BLOSC_STRICT_ALIGN)
   uint16_t chunk;
   memcpy(&chunk, from, 2);
   memcpy(out, &chunk, 2);
-#else
-  *(uint16_t *) out = *(uint16_t *) from;
-#endif
   return out + 2;
 }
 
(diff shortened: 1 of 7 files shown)


thesamesam:
cc @mgorny @maskray @soapgentoo

MaskRay:
Thanks! This gets rid of the conditions that are very difficult to get right:

```
/*
 * Detect if the architecture is fine with unaligned access.
 */
#if !defined(BLOSC_STRICT_ALIGN)
#define BLOSC_STRICT_ALIGN
#if defined(__i386__) || defined(__386) || defined (__amd64)  /* GNU C, Sun Studio */
#undef BLOSC_STRICT_ALIGN
#elif defined(__i486__) || defined(__i586__) || defined(__i686__)  /* GNU C */
#undef BLOSC_STRICT_ALIGN
#elif defined(_M_IX86) || defined(_M_X64)   /* Intel, MSVC */
#undef BLOSC_STRICT_ALIGN
#elif defined(__386)
#undef BLOSC_STRICT_ALIGN
#elif defined(_X86_) /* MinGW */
#undef BLOSC_STRICT_ALIGN
#elif defined(__I86__) /* Digital Mars */
#undef BLOSC_STRICT_ALIGN
/* Modern ARM systems (like ARM64) should support unaligned access
   quite efficiently. */
#elif defined(__ARM_FEATURE_UNALIGNED)   /* ARM, GNU C */
#undef BLOSC_STRICT_ALIGN
#elif defined(_ARCH_PPC) || defined(__PPC__)
/* Modern PowerPC systems (like POWER8) should support unaligned access
   quite efficiently. */
#undef BLOSC_STRICT_ALIGN
#endif
#endif
```

https://github.com/riscv-non-isa/riscv-c-api-doc/pull/49 says "slow misaligned accesses may be very slow ". The long list would also cause continued problems with new architectures.

blosc/fastcopy.c
```diff
@@ -40,13 +38,9 @@ static inline unsigned char *copy_1_bytes(unsigned char *out, const unsigned cha
 }
 
 static inline unsigned char *copy_2_bytes(unsigned char *out, const unsigned char *from) {
-#if defined(BLOSC_STRICT_ALIGN)
   uint16_t chunk;
   memcpy(&chunk, from, 2);
   memcpy(out, &chunk, 2);
```
DerDakon, line 43: Why does this need an intermediate variable? `memcpy()` should handle this fine.

blosc/fastcopy.c
```diff
@@ -56,13 +50,9 @@ static inline unsigned char *copy_3_bytes(unsigned char *out, const unsigned cha
 }
 
 static inline unsigned char *copy_4_bytes(unsigned char *out, const unsigned char *from) {
-#if defined(BLOSC_STRICT_ALIGN)
   uint32_t chunk;
   memcpy(&chunk, from, 4);
   memcpy(out, &chunk, 4);
```
DerDakon, line 55: intermediate variable?

blosc/fastcopy.c
```diff
@@ -99,18 +85,13 @@ static inline unsigned char *copy_16_bytes(unsigned char *out, const unsigned ch
...
-   int i;
-   for (i = 0; i < 16; i++) {
-     *out++ = *from++;
-   }
-#endif
+  int i;
+  for (i = 0; i < 16; i++) {
+    *out++ = *from++;
+  }
   return out;
```
DerDakon, line 93: This is just a `memcpy()` as well. Given that compilers may already replace that loop with a call to memcpy() anyway just use that.

blosc/fastcopy.c
```diff
@@ -127,15 +108,6 @@ static inline unsigned char *copy_32_bytes(unsigned char *out, const unsigned ch
...
-  from += 8; out += 8;
-  *(uint64_t*)out = *(uint64_t*)from;
-  from += 8; out += 8;
-  *(uint64_t*)out = *(uint64_t*)from;
-  from += 8; out += 8;
-  *(uint64_t*)out = *(uint64_t*)from;
-  from += 8; out += 8;
 #else
   int i;
   for (i = 0; i < 32; i++) {
```
DerDakon, line 113: use use `memcpy()`

FrancescAlted:
> Modern compilers are capable of optimising the "slow" path bitshifts and memcpy into faster alternatives where it is legal.

Apparently this is not the case with at least MacOS with M1 and clang 14.0.3 (and I remember similar results with Intel/Linux).  Using the bench suite for MacOS on M1, and using unaligned access (did several runs and kept the best result):

```
$  bench/b2bench blosclz shuffle single 8 8388608 4 19
<snip>
Round-trip compr/decompr on 2.5 GB
Elapsed time:	    0.3 s, 17036.6 MB/s
```

and with your PR (same,  did several runs and kept the best result):

```
$  bench/b2bench blosclz shuffle single 8 8388608 4 19
<snip>
Round-trip compr/decompr on 2.5 GB
Elapsed time:	    0.5 s, 12415.2 MB/s
```

So, unaligned access is still 1.37x  faster.

thesamesam:
On Linux 6.4.11 with AMD Zen 2 (3950x) on tmpfs with gcc 13.2.1 20230819 and clang 16.0.6:
```console
Benchmark 1: build-gcc-generic-vanilla/bench/b2bench blosclz shuffle single 8 8388608 4 19
  Time (mean ± σ):     421.0 ms ±   4.0 ms    [User: 1124.5 ms, System: 240.3 ms]
  Range (min … max):   412.7 ms … 430.3 ms    100 runs

Benchmark 2: build-gcc-generic-new/bench/b2bench blosclz shuffle single 8 8388608 4 19
  Time (mean ± σ):     425.0 ms ±   3.7 ms    [User: 1160.9 ms, System: 242.4 ms]
  Range (min … max):   417.4 ms … 436.3 ms    100 runs

Summary
  'build-gcc-generic-vanilla/bench/b2bench blosclz shuffle single 8 8388608 4 19' ran
    1.01 ± 0.01 times faster than 'build-gcc-generic-new/bench/b2bench blosclz shuffle single 8 8388608 4 19'
```

```console
Benchmark 1: build-gcc-O3-vanilla/bench/b2bench blosclz shuffle single 8 8388608 4 19
  Time (mean ± σ):     421.7 ms ±   4.3 ms    [User: 1121.6 ms, System: 243.5 ms]
  Range (min … max):   413.0 ms … 430.0 ms    100 runs

Benchmark 2: build-gcc-O3-new/bench/b2bench blosclz shuffle single 8 8388608 4 19
  Time (mean ± σ):     423.3 ms ±   4.0 ms    [User: 1122.0 ms, System: 244.7 ms]
  Range (min … max):   413.6 ms … 440.0 ms    100 runs

Summary
  'build-gcc-O3-vanilla/bench/b2bench blosclz shuffle single 8 8388608 4 19' ran
    1.00 ± 0.01 times faster than 'build-gcc-O3-new/bench/b2bench blosclz shuffle single 8 8388608 4 19'
```

```console
Benchmark 1: build-clang-generic-vanilla/bench/b2bench blosclz shuffle single 8 8388608 4 19
  Time (mean ± σ):     420.3 ms ±   3.9 ms    [User: 1104.9 ms, System: 240.5 ms]
  Range (min … max):   411.3 ms … 426.0 ms    100 runs

Benchmark 2: build-clang-generic-new/bench/b2bench blosclz shuffle single 8 8388608 4 19
  Time (mean ± σ):     415.3 ms ±   4.3 ms    [User: 1065.3 ms, System: 245.5 ms]
  Range (min … max):   406.8 ms … 422.7 ms    100 runs

Summary
  'build-clang-generic-new/bench/b2bench blosclz shuffle single 8 8388608 4 19' ran
    1.01 ± 0.01 times faster than 'build-clang-generic-vanilla/bench/b2bench blosclz shuffle single 8 8388608 4 19'
```

So, on Linux/AMD, it's within noise.

On macOS 13.5, Intel i9-9880H, Apple clang version 14.0.3 (clang-1403.0.22.14.1), the variation is terrible (and the result is very inconsistent, swinging in both directions) and I suspect it's because of throttling. I've asked someone if I can try on their arm mac mini.

FrancescAlted:
> Using this:
> ...
> So, on Linux/AMD, it's within noise.

This is interesting.  With Linux/Intel I still see a difference (see later).

> On macOS 13.5, Intel i9-9880H, Apple clang version 14.0.3 (clang-1403.0.22.14.1), the variation is terrible (and the result is very inconsistent, swinging in both directions) and I suspect it's because of throttling. I've asked someone if I can try on their arm mac mini.

What could be happening here is that threading makes throttling to affect measurements a lot.  For this benchmark, I think it is better to restrict oneself to 1 single thread.  With that, on MacOS 13.4.1, M1 and clang 14.0.3, and unaligned reads:

```
$ bench/b2bench blosclz shuffle single 1 8388608 4 19
<snip>
Round-trip compr/decompr on 2.5 GB
Elapsed time:	    0.8 s, 6682.5 MB/s
```

Whereas with the current version of this PR:

```
$ bench/b2bench blosclz shuffle single 1 8388608 4 19
<snip>
Round-trip compr/decompr on 2.5 GB
Elapsed time:	    1.3 s, 4463.2 MB/s
```

So in this case, unaligned reads makes the compression operation 1.5x faster.

For an Intel i9-13900K / Linux 6.2.0 (Ubuntu  23.04), gcc 12.3.0 and unaligned reads:

```
$ bench/b2bench blosclz shuffle single 1 8388608 4 19
<snip>
Round-trip compr/decompr on 2.5 GB
Elapsed time:	    0.6 s, 9758.3 MB/s
```

With the current version of this PR:

```
$ bench/b2bench blosclz shuffle single 1 8388608 4 19
<snip>
Round-trip compr/decompr on 2.5 GB
Elapsed time:	    0.6 s, 9064.7 MB/s
```

In this case the speed-up is less, but still, unaligned reads produces non-negligible faster compression operation (1.07x).

MaskRay:
`memcpy(dst, src, 4);` has been optimized by GCC for 10+ years (check godbolt 4.1 from 2007), if not 20+/30+ years.

This one can not be optimized by GCC or Clang: `#define NDLZ_READU32(p) ((p)[0] | (p)[1]<<8 | (p)[2]<<16 | (p)[3]<<24)`. It may need a switch to `memcpy`, if this is a possible bottleneck.

MaskRay:
I can reproduce `bench/b2bench blosclz shuffle single 1 8388608 4 19` slowdown with this patch, but only with GCC, not Clang.

I create two build directories for main (out/gcc0) and no-unaligned-access-ub (out/gcc1).
```
% cat =cmp-objdump
#!/bin/zsh
nvim -d =(/tmp/Rel/bin/llvm-objdump -d --symbolize-operands -M intel --no-leading-addr --no-show-raw-insn $1 --disassemble-symbols=$(</tmp/0)) =(/tmp/Rel/bin/llvm-objdump -d --symbolize-operands -M intel --no-leading-addr --no-show-raw-insn $2 --disassemble-symbols=$(</tmp/0))
% cmp-objdump out/gcc[01]/bench/b2bench
```
`out/gcc1/bench/b2bench` has larger `blosclz_compress` due to GCC unoptimized `#define NDLZ_READU32(p) ((p)[0] | (p)[1]<<8 | (p)[2]<<16 | (p)[3]<<24)`
```
                movzx   esi, byte ptr [rax - 0x2]
                movzx   edi, byte ptr [rax - 0x1]
                sub     rdx, rbx
                add     r9d, 0x1
                xor     r8d, r8d
                shl     edi, 0x10
                shl     esi, 0x8
                or      esi, edi
                movzx   edi, byte ptr [rax - 0x3]
                or      esi, edi
                movzx   edi, byte ptr [rax]
                shl     edi, 0x18
                or      esi, edi
                imul    esi, esi, 0x9e3779b1
                shr     esi, cl
```

I suggest that we replace `NDLZ_READU32` uses with `memcpy`.

blosc/fastcopy.c
MaskRay: IMHO this clean up does not need to be done in this patch. blosc2 maintainers are probably in a better position to follow up and fix these:)

But I don't mind if thesamesam has cleaned up this part as well:)

thesamesam:
I'll play around with this some more during the week and let you know if/when it needs re-review. Thanks!

MaskRay:
If it helps, creating two patches may be feasible.

One is to replace things like `*(uint64_t*)out = *(uint64_t*)from;` with `memcpy`. This is always fine for optimizations and usually evens leads to identical output file.

The other is on `#define NDLZ_READU32(p) ((p)[0] | (p)[1]<<8 | (p)[2]<<16 | (p)[3]<<24)` and/or removing the `BLOSC_STRICT_ALIGN` setting.

The idea is that the first patch is entirely safe and can be merged immediately, while the second needs performance evaluation.

FrancescAlted:
Any news here?

FrancescAlted:
Closing due to inactivity

thesamesam:
Yeah, sorry I haven't come back to this yet. I'll try to come back to it but I've not had a chance. Thanks.

PR #551, opened 2023-08-21
Disable visibility attribute for mingw

Biswa96:
The same thing is done with BLOSC_EXPORT macro also.

    This fixes the following compiler warning with mingw gcc for static build.

    shuffle-generic.c: In function 'shuffle_generic':
    shuffle-generic.c:18:1: warning: visibility attribute not supported
    in this configuration; ignored [-Wattributes]


FrancescAlted:
Looks good!  Thanks.

PR #561, opened 2023-09-21
Fix typo newly found by codespell

DimitriPapadopoulos:


DimitriPapadopoulos:
The `CMake Error: CMake was unable to find a build program corresponding to "Ninja"` CI error is unrelated.

PR #562, opened 2023-10-03
Make tuners return error

martaiborra:
- Make tuners return error 
- Check if blosc2_context has been created correctly


FrancescAlted (approved):
LGTM!

PR #565, opened 2023-10-26
Expose b2nd_copy_buffer function

ivilata:
This exposes the utility function `b2nd_copy_buffer`, which may be useful for other projects dealing with multidimensional arrays in memory. Documentation and tests are included.


FrancescAlted (approved):
Looks good to me!

PR #567, opened 2023-10-26
New bitshuffle functions

FrancescAlted:
Here it is a new implementation of bitshuffle for AVX512 coming from [upstream](https://github.com/kiyo-masui/bitshuffle).  I also took the opportunity to sync the SSE2, AVX2 and ARM/NEON implementions (however, the latter is still pretty much slower than the generic solution, so it is not enabled by default).

Preliminary results using AVX512 on Zen4 (AMD 7950X) points to an speed-up of 10% to 15%, which is quite good (most specially as [Zen4 does not have a full 512-bit register implementation](https://www.anandtech.com/show/21092/amd-unveils-ryzen-threadripper-7000-family-zen-4-for-workstations-and-hedt)).

Finally, AVX512 codepath will be compiled by default on UNIX platforms (still need to figure out which MSVC version supports AVX512), and will be used when AVX512 (more specifically AVX512F *and* AVX512BW) presence should be detected in runtime.

PR #568, opened 2023-10-27
Add files for GNU Guix support

ivilata:
This adds the files needed to ease development and installation from upstream source under GNU Guix, as described in [From development environments to continuous integration](https://guix.gnu.org/es/blog/2023/from-development-environments-to-continuous-integrationthe-ultimate-guide-to-software-development-with-guix/). For instance, spawning a shell with build dependencies would be a easy as `guix shell -CP -D -f guix.scm`, and building just `guix build -f guix.scm`. Variants supporting AVX2 and AVX-512 (for #567) are provided too (`guix -L $PWD/.guix/modules c-blosc2-with-avx2`), as some packages in GNU Guix disable it by default, most notably `c-blosc`.

### DIFF
--- a/.guix-channel
+++ b/.guix-channel
@@ -1,0 +1,3 @@
+(channel
+  (version 0)
+  (directory ".guix/modules"))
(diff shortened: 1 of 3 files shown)


FrancescAlted (approved):
LGTM

PR #571, opened 2023-11-09
Use typedef for blosc_timestamp_t

mulimoen:
Using a `typedef` makes code analysis more efficient as one can see `blosc_timestamp_t` is a type straigth away. In addition tools such as [`bindgen`](https://rust-lang.github.io/rust-bindgen/) are able to generate better bindings.


FrancescAlted:
Thanks @mulimoen !

PR #576, opened 2024-01-23
Initialize grok lib when loading the library

martaiborra:
This PR is needed for initializing the grok library if the blosc2-grok plugin is used without calling `import blosc2_grok`.


FrancescAlted (approved):
LGTM

PR #577, opened 2024-01-23
Implement a new truncation precision for integers (BLOSC_FILTER_INT_TRUNC)

FrancescAlted:
We already have a `TRUNC` filter, but this only is good for `float32_t` and `float64_t`.  This new `BLOSC_FILTER_INT_TRUNC` is for truncating integers to a given number of bits.  This is useful for compressing integers that are not using all the bits of the type.


martaiborra (approved):
LGTM!

PR #578, opened 2024-01-24
Optimizing zstd, specially when using dicts

FrancescAlted:
This is an effort to optimize the dictionaries for Zstd inside Blosc2.

The tuning here is mainly based on experiments done with `test_dict_schunk.c`, but that is probably very dependent on data.

Also, during testing, I detected that splitting is also beneficial for low compression levels in Zstd, so it has been enable for it when using SHUFFLE and clevel <= 5.

### DIFF
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -1755,8 +1755,7 @@ static int blosc_d(
   }
 
   /* The number of compressed data streams for this block */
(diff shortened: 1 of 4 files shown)


FrancescAlted:
If tests pass, this is meant to be merged into main asap.

martaiborra (approved):
LGTM!

PR #580, opened 2024-01-26
Adding decompress_file example

dmikushin:
The `compress_file` example nicely demonstrates the Blosc2 compression procedure. It would be highly educative though to include also a simple example, which performs the inverse operation. Therefore, this PR adds `decompress_file` example, in order to have the complete loop of compression/decompression functionality at glance.


FrancescAlted:
Cool!  Thanks @dmikushin

dmikushin:
Thanks for fixing my issues, I've noted what I did wrong.

PR #588, opened 2024-02-12
Change blosc2.c:set_values to always call memcpy if BLOSC_STRICT_ALIGN

mwhudson:
While I have some sympathy with the argument in #550 that all this code depending on unaligned access should go away in favour of letting the compiler put it back when it makes sense, this particular instance of it is all that is making the test suite fail on 32-bit ARM on the current Ubuntu devel series.


FrancescAlted:
It seems a bit strange to me that this simple PR is the only thing required to make 32-bit ARM tests happy on Ubuntu, but if that is the case, I have nothing against it :-). Thanks @mwhudson !

PR #589, opened 2024-02-21
No shuffle repeat

FrancescAlted:
We introduced the capability of repeating the shuffle filter (and only the shuffle) in C-Blosc2, but we recently discovered issues when using repeat values that are odd (i.e. not even).  This PR removes this capability because:

1) It is buggy
2) It was not documented
3) In our tests, it does not provide any significant advantage in compression ratio
4) It complicates code

As it was not documented (besides than buggy), we don't expect this would break anybody's code.

Also, this PR completes documentation on internal filters (essentially, it explicitly says that `filters_meta` does not have any effect on shuffle, bitshuffle and delta filters).


martaiborra (approved):
LGTM!

Issue #533, opened 2023-07-05
Building with latest oneAPI intel compilers COMPILER_ID == IntelLLVM


zerothi:
**Describe the bug**
The compilation and detection of the intel oneAPI compiler fails.

**To Reproduce**

```
CC=icx cmake ...
```
fails.

The problem arises since the compiler ID is shown to be IntelLLVM which breaks nearly all of the COMPILER_ID checks present in the CMakeLists file.
```
-- Building for compiler ID IntelLLVM
```
It would be great if this could be amended in some way.
All newer Intel compilers will be based on LLVM and the COMPILER ID will be: `IntelLLVM`.

**System information:**
 - Intel 2023.0.1

**Additional context**
I believe the errors this produce
```
/zhome/0e/2/36189/dcc/bash-build/XeonGold6126/intel/c-blosc2-2.8.0/plugins/filters/bytedelta/bytedelta.c:32:54: error: '__builtin_ia32_palignr128' needs target feature ssse3
```
are only due to the cmake files not being able to detect the features allowed by the compiler. So I think it should be solveable in the cmake files.

Let me know if you need anything else.

FrancescAlted:
Hi there.  Sorry for the late reply.  We don't have resources for supporting compilers beyond gcc/clang and MSVC, but we are receptive to accept PRs for new ones.  If interested, please send yours.

FrancescAlted:
I have merged this, and a new release (2.14.0) will be out soon.

PR #595, opened 2024-03-19
changed STREQUAL to MATCHES

zerothi:
This changes the logic in the CMake infrastructure somewhat.

STREQUAL is an exact match.
MATCHES just matches a string in the full string.
E.g.

  ... MATCHES ABC

will find for "fooABCbar" and all variants where ABC is present. This may sometimes not be decired.
However, in the case for compilers and systems, using matches may be the best choice to allow smaller variations of variables (coming down the road).

In this change SYSTEM_PROCESSOR will now use MATCHES instead of STREQUAL.
The original code could be dated back to 2015 in
1c36bf150.

The same goes for the compiler specifications.
Now MATCHES allows a greater variability in the
compiler naming conventions. This change results
in some logic changes (should be changed if not
desired).

For instance this change:

-    elseif(CMAKE_C_COMPILER_ID STREQUAL Clang AND CMAKE_C_COMPILER_VERSION VERSION_GREATER 13)
+    elseif(CMAKE_C_COMPILER_ID MATCHES Clang AND CMAKE_C_COMPILER_VERSION VERSION_GREATER 13)

results in ALTIVEC=TRUE for AppleClang compilers at versions >=13. It did not previously.

While MATCHES Intel will also match IntelLLVM I have added explicit IntelLLVM to make it clear the
intent, this is strictly not needed due to the
above.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -271,16 +271,12 @@ endif()
 # for newer hardware on older machines as well as cross-compilation.
 message(STATUS "Building for system processor ${CMAKE_SYSTEM_PROCESSOR}")
 message(STATUS "Building for compiler ID ${CMAKE_C_COMPILER_ID}")
-if(CMAKE_SYSTEM_PROCESSOR STREQUAL i386 OR
-    CMAKE_SYSTEM_PROCESSOR STREQUAL i686 OR
-    CMAKE_SYSTEM_PROCESSOR STREQUAL x86_64 OR
-    CMAKE_SYSTEM_PROCESSOR STREQUAL amd64 OR
-    CMAKE_SYSTEM_PROCESSOR STREQUAL AMD64)
-    if(CMAKE_C_COMPILER_ID STREQUAL GNU)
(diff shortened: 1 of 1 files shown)


zerothi:
This should solve #533, but I haven't tested it yet!

Would something like this be applicable?

FrancescAlted:
> This should solve #533, but I haven't tested it yet!
> 
> Would something like this be applicable?

Indeed.  Keep us informed if this is enough and I will be happy to merge.  Thanks!

KingKarp4 (approved):

KingKarp4 (approved):

zerothi:
Thanks!

FYI, I just tried this with Intel oneAPI 2024.0.1.46, and the built succeeds, though with some warnings related to unknown arguments when building zlib-ng.

Other than that, everything runs smoothly. All 1745 tests pass! :)

KingKarp4:
Thanx for the updates on your own time in the future and you will find the
perfect way for your students to know what they need to be in the right
position for the first time investment goals 👣


MARCEL.

PR #603, opened 2024-04-24
Fix musl linux build

### LINKED ISSUE #602: Feature Request: Opt ouf of `blosc_get_cpu_features`
Hello there.

I'm trying to build on musllinux, and also had this issue on x86_64-apple-darwin when statically linking:

```
/usr/local/musl/bin/../lib/gcc/x86_64-unknown-linux-musl/11.2.0/../../../../x86_64-unknown-linux-musl/bin/ld: /tmp/rustc88VirH/libblosc2-d98c3a018c009f3b.rlib(shuffle.c.o): in function `bitunshuffle':
          shuffle.c:(.text+0x523): undefined reference to `__cpu_model'
```

full example from workflow: https://github.com/milesgranger/cramjam/actions/runs/8749803819/job/24012143129#step:10:361

---

It appears this is based off determining `HAVE_CPU_FEAT_INTRIN` here: https://github.com/Blosc/c-blosc2/blob/7611351ad7f9740046dfaeb15ea4f748da2d6bad/blosc/shuffle.c#L57

Then will try to get cpu features here:
https://github.com/Blosc/c-blosc2/blob/7611351ad7f9740046dfaeb15ea4f748da2d6bad/blosc/shuffle.c#L96

---

### Feature request

I've found if I add an additional logic to opt out, ex "OPT_OUT_CPU_FEAT" to get this function instead, then everything is happy again. But there is also a good chance I'm missing something here. :)
https://github.com/Blosc/c-blosc2/blob/7611351ad7f9740046dfaeb15ea4f748da2d6bad/blosc/shuffle.c#L290

milesgranger:
Will close #602 

Seems this is the idiom to check for musl libc and [appears happy on blosc2-rs' end](https://github.com/milesgranger/blosc2-rs/actions/runs/8814078348/job/24193177277) using this branch.

Not sure if you want the added trouble of a musl build in your CI as well, but I'll be happy to complain if mine breaks down the line. ;-)

### DIFF
--- a/blosc/shuffle.c
+++ b/blosc/shuffle.c
@@ -96,6 +96,8 @@ typedef enum {
 #if defined(HAVE_CPU_FEAT_INTRIN)
 static blosc_cpu_features blosc_get_cpu_features(void) {
   blosc_cpu_features cpu_features = BLOSC_HAVE_NOTHING;
+
+  __builtin_cpu_init();
   if (__builtin_cpu_supports("sse2")) {
     cpu_features |= BLOSC_HAVE_SSE2;
   }


FrancescAlted:
It looks like there are several failures in CI.  I am not too worried about the builds for Ubuntu (temporary failure in upstream repos?), but the one with fuzzer seems significant: https://github.com/Blosc/c-blosc2/actions/runs/8814267732/job/24194115220?pr=603

Would you be able to see what it is happening there?

milesgranger:
Interesting, I should have some time tomorrow to try and figure it out. :+1:

milesgranger:
Alrighty, it appears adding logic to identify musl is [bad practice](https://openwall.com/lists/musl/2013/03/29/13) which after some digging found that to resolve the undefined `__cpu_model` refs is to link libgcc. I think technically...maybe, this ought to be in c-blosc2 configuration somewhere if the target env is musl but I'm happy [to configure this in the rust bindings](https://github.com/milesgranger/blosc2-rs/pull/10/commits/50a86f3343a5507dbff2ab4f0eaeb68a7601db9f). 

I added one commit here to call `__builtin_cpu_init()` as I thought initially this was needed from [reading the docs](https://gcc.gnu.org/onlinedocs/gcc/x86-Built-in-Functions.html#index-_005f_005fbuiltin_005fcpu_005finit-1), but I think it's not needed given the caveat there:

> ...only when used in a function that is executed before any constructors are called. The CPU detection code is automatically executed in a very high priority constructor. 

So I think there is nothing to do here now, closing it. Thanks for the help. :)

FrancescAlted:
Sounds good.  FWIW, we are building [binary wheels for blosc2 and musl](https://pypi.org/project/blosc2/#files), including binaries for libs and include files:

```
faltet@Mac-mini-de-Francesc ~/b/c-blosc2 (main)> python -c "import blosc2; print(blosc2.__path__)"       (python-blosc2)
['/Users/faltet/miniconda3/envs/python-blosc2/lib/python3.12/site-packages/blosc2']
faltet@Mac-mini-de-Francesc ~/b/c-blosc2 (main)> ls /Users/faltet/miniconda3/envs/python-blosc2/lib/libblosc2*
/Users/faltet/miniconda3/envs/python-blosc2/lib/libblosc2.2.14.3.dylib*
/Users/faltet/miniconda3/envs/python-blosc2/lib/libblosc2.2.dylib*
/Users/faltet/miniconda3/envs/python-blosc2/lib/libblosc2.a
/Users/faltet/miniconda3/envs/python-blosc2/lib/libblosc2.dylib*
faltet@Mac-mini-de-Francesc ~/b/c-blosc2 (main)> ls /Users/faltet/miniconda3/envs/python-blosc2/include/blosc2*
/Users/faltet/miniconda3/envs/python-blosc2/include/blosc2.h

/Users/faltet/miniconda3/envs/python-blosc2/include/blosc2:
blosc2-common.h     blosc2-export.h     blosc2-stdio.h      codecs-registry.h   filters-registry.h  tuners-registry.h
```

so having access to a development environment for Blosc2 is just a `pip install blosc2` away.

PR #606, opened 2024-05-12
Fix trailer_len sizing mismatch

milesgranger:
In `frame_update_trailer` the trailer len is `int64_t` which when written to the trailer len is expected to be `uint32_t`. On big-endian systems, this causes subsequent reads of cframes to have incorrect sizing as `frame_from_cframe` uses the correct 32bit size.

This will fix https://github.com/milesgranger/blosc2-rs/issues/23 for me, not sure how / if you want this tested on s390x / big endian.

Somewhat related to https://github.com/Blosc/c-blosc2/issues/467, but there as I understand it is when using a file backed schunk.

### DIFF
--- a/blosc/frame.c
+++ b/blosc/frame.c
@@ -551,7 +551,7 @@ int frame_update_trailer(blosc2_frame_s* frame, blosc2_schunk* schunk) {
   }
 
   // Create the trailer in msgpack (see the frame format document)
-  int64_t trailer_len = FRAME_TRAILER_MINLEN;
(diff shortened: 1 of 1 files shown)


FrancescAlted:
Looks good to me.  Thanks @milesgranger !

mr-c:
Applying this PR on top of https://github.com/Blosc/c-blosc2/releases/tag/v2.14.4 did not result in any additional tests passing on s390x for me (still 90 tests fail)

PR #607, opened 2024-05-12
Add functions for c|dparams, storage and io defaults

milesgranger:
Hello there,

While working on the rust bindings to c-blosc2 I currently zero initialize the C|DParams and Storage structs and works I think mostly by accident. I do this because the suggested use of `BLOSC2_STORAGE_DEFAULTS` and `BLOSC2_C|DPARAMS_DEFAULTS` doesn't work due to a [different bug in bindgen](https://github.com/rust-lang/rust-bindgen/issues/1888). 

I can understand if you don't want to include this, but it'd help me avoid maintaining a branch for compatibility. :) If you do, I'm happy to add/expand the docs to these functions.

### DIFF
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -4695,10 +4695,16 @@ int blosc2_get_slice_nchunks(blosc2_schunk* schunk, int64_t *start, int64_t *sto
 
 blosc2_cparams blosc2_get_blosc2_cparams_defaults(void) {
   return BLOSC2_CPARAMS_DEFAULTS;
-}
+};
+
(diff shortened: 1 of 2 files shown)


FrancescAlted:
Well, while I am somewhat reluctant to add unnecessary API calls, I'd like to support your efforts on the rust front, so let's do it.  Also, for completeness, could you add the corresponding API for `BLOSC2_IO_DEFAULTS` too?  This should be useful [soon](https://github.com/Blosc/c-blosc2/pull/604).

FrancescAlted:
Thanks @milesgranger !

PR #612, opened 2024-05-31
Use mode parameter only in open call

JanSellner:
This PR is an intend to fix the problem mentioned in [https://github.com/Blosc/python-blosc2/pull/198](https://github.com/Blosc/python-blosc2/pull/198) of occasional bus errors. I have not encountered further bus errors after this fix.

After this, we can start benchmarking in Python.


FrancescAlted (approved):
LGTM

Issue #430, opened 2022-12-14
Build issue with `universal2` on macos


t20100:
**Describe the bug**
On macos, building a `universal2` build  (`-arch arm64 -arch x86_64` options) of `c-blosc2` fails when enabling `SHUFFLE_[SSE2/AVX2]_ENABLED` as those options are activated both for `x86_64` and `arm64`. I expect similar issue the other way around with `SHUFFLE_NEON_ENABLED`.

Related to similar issue on `c-blosc`: https://github.com/Blosc/c-blosc/issues/334

**To Reproduce**
On a macos ARM computer with a `universal2` build of Python, compile `blosc2` HDF5 filter as part of `hdf5plugin` (blosc2 PR: https://github.com/silx-kit/hdf5plugin/pull/201) with `python setup.py build`

**Expected behavior**
Even if `SHUFFLE_[SSE2/AVX2/NEON]_ENABLED` are set, if the target platform does not support those instructions they should not be considered.
I would expect the same behavior as proposed for blosc(1) in PR https://github.com/Blosc/c-blosc/pull/347.

**Logs**

<details><summary>Compilation logs...</summary>
```
gcc -Wno-unused-result -Wsign-compare -Wunreachable-code -fno-common -dynamic -DNDEBUG -g -fwrapv -O3 -Wall -arch arm64 -arch x86_64 -g -DHAVE_LZ4=1 -DHAVE_ZLIB=1 -DHAVE_ZSTD=1 -DH5_USE_18_API -DSHUFFLE_SSE2_ENABLED=1 -Isrc/hdf5/include -Isrc/hdf5/include/darwin -Isrc/c-blosc2 -Isrc/c-blosc2/blosc -Isrc/c-blosc2/include -Isrc/c-blosc/internal-complibs/lz4-1.9.3 -Isrc/c-blosc/internal-complibs/zlib-1.2.11 -Isrc/c-blosc/internal-complibs/zstd-1.5.2 -Isrc/c-blosc/internal-complibs/zstd-1.5.2/common -Isrc/PyTables/hdf5-blosc2/src -Ivenv/py39/include -I/Library/Frameworks/Python.framework/Versions/3.9/include/python3.9 -c src/c-blosc2/blosc/bitshuffle-sse2.c -o build/temp.macosx-10.9-universal2-3.9/src/c-blosc2/blosc/bitshuffle-sse2.o -std=gnu99 -O3 -ffast-math -pthread -mcpu=native
[...]

PR #622, opened 2024-06-25
Do not pass `-m` flags when compiling `shuffle.c`

### LINKED ISSUE #621: `shuffle.c` is compiled with `-mavx512*` which results in `SIGILL` on lower CPUs
**Describe the bug**
When compiler supports SSE2/AVX2/AVX512 instructions, the respective `-m` flags are added to the flags for `shuffle.c`. As a result, the compiler can compile arbitrary code with these instruction sets. For me, GCC 14 creates a library that uses AVX512 instructions on unsupported CPU and effectively crashes with SIGILL, e.g.:

```
$ ninja test
[0/1] Running tests...
Test project /tmp/c-blosc2/build
          Start    1: test_plugin_test_ndlz
   1/1746 Test    #1: test_plugin_test_ndlz .....................................***Exception: Illegal  0.11 sec
          Start    2: test_plugin_test_zfp_acc_float
   2/1746 Test    #2: test_plugin_test_zfp_acc_float ............................***Exception: Illegal  0.10 sec
          Start    3: test_plugin_test_zfp_prec_float
```

For example, `plugins/codecs/ndlz/test_ndlz` crashes in `get_shuffle_implementation`:

```
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│   0x555555582115 <get_shuffle_implementation+37>  jne    0x5555555821d0 <get_shuffle_implementation+224>                            │
│   0x55555558211b <get_shuffle_implementation+43>  vmovq  0x10ae55(%rip),%xmm0        # 0x55555568cf78                               │
│   0x555555582123 <get_shuffle_implementation+51>  vmovq  0x10adad(%rip),%xmm1        # 0x55555568ced8                               │
│   0x55555558212b <get_shuffle_implementation+59>  lea    0xe250a(%rip),%rcx        # 0x55555566463c                                 │
│   0x555555582132 <get_shuffle_implementation+66>  lea    0xa2807(%rip),%rax        # 0x555555624940 <unshuffle_generic>             │
│   0x555555582139 <get_shuffle_implementation+73>  lea    0xa2fc0(%rip),%rdx        # 0x555555625100 <bshuf_untrans_bit_elem_scal>   │
│   0x555555582140 <get_shuffle_implementation+80>  vpinsrq $0x1,%rdx,%xmm1,%xmm1                                                     │
│   0x555555582146 <get_shuffle_implementation+86>  vpinsrq $0x1,%rax,%xmm0,%xmm0                                                     │
│   0x55555558214c <get_shuffle_implementation+92>  mov    %rcx,(%rdi)                                                                │
│   0x55555558214f <get_shuffle_implementation+95>  mov    %rdi,%rax                                                                  │
│  >0x555555582152 <get_shuffle_implementation+98>  vinserti32x4 $0x1,%xmm1,%ymm0,%ymm0                                               │
│   0x555555582159 <get_shuffle_implementation+105> vmovdqu %ymm0,0x8(%rdi)                                                           │
│   0x55555558215e <get_shuffle_implementation+110> vzeroupper                                                                        │
│   0x555555582161 <get_shuffle_implementation+113> ret                                                                               │
│   0x555555582162 <get_shuffle_implementation+114> data16 cs nopw 0x0(%rax,%rax,1)                                                   │
│   0x55555558216d <get_shuffle_implementation+125> nopl   (%rax)                                                                     │
│   0x555555582170 <get_shuffle_implementation+128> cmp    $0x208000,%edx                                                             │
│   0x555555582176 <get_shuffle_implementation+134> je     0x5555555821a0 <get_shuffle_implementation+176>                            │
│   0x555555582178 <get_shuffle_implementation+136> vmovq  0x10ad68(%rip),%xmm0        # 0x55555568cee8                               │
│   0x555555582180 <get_shuffle_implementation+144> vmovq  0x10ae30(%rip),%xmm1        # 0x55555568cfb8                               │
└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
multi-thre Thread 0x7ffff7caa7 In: get_shuffle_implementation                                                 L306  PC: 0x555555582152 
```

**To Reproduce**
```
cmake -G Ninja ${srcdir} -DCMAKE_C_FLAGS='-march=znver2 -O2 -pipe' -DCMAKE_BUILD_TYPE=RelWithDebInfo
ninja
ninja test
```

**Expected behavior**
Not crashing, i.e. not using SSE2/AVX2/AVX512 instructions outside specific files intended for them :-).

**Logs**
CMake output: [cmake.txt](https://github.com/user-attachments/files/15931082/cmake.txt)
Ninja output (for build): [build.txt](https://github.com/user-attachments/files/15931083/build.txt)

**System information:**
 - OS: Gentoo Linux amd64
 - Compiler: GCC 14.1.1
 - Version: 2.15.0

**Additional context**
n/a

mgorny:
Do not pass `-msse2`, `-mavx2`, etc. flags to the compiler when compiling `shuffle.c`.  From what I can see, the file itself does not use any of these intrinsics, and they are only used by functions declared in `bitshuffle-*.c` and `shuffle-*.c` (where the respective flags are still passed).  This prevents the compiler from incidentally optimizing the code called independenlty of the runtime CPU check to these instruction sets, effectively causing `SIGILL` on other CPUs.

I have verified that this fixes the issue on `-march=znver2`, but also does not cause any issues on `-march=x86-64` and `-march=i686`.

Fixes #621

### DIFF
--- a/blosc/CMakeLists.txt
+++ b/blosc/CMakeLists.txt
@@ -300,17 +300,11 @@ if(COMPILER_SUPPORT_SSE2)
             set_source_files_properties(
                     shuffle-sse2.c bitshuffle-sse2.c blosclz.c fastcopy.c
                     PROPERTIES COMPILE_OPTIONS "/arch:SSE2")
-            set_property(
-                    SOURCE shuffle.c
-                    APPEND PROPERTY COMPILE_OPTIONS "/arch:SSE2")
         endif()
     else()
         set_source_files_properties(
                 shuffle-sse2.c bitshuffle-sse2.c blosclz.c fastcopy.c
                 PROPERTIES COMPILE_OPTIONS -msse2)
-        set_property(
-                SOURCE shuffle.c
-                APPEND PROPERTY COMPILE_OPTIONS -msse2)
         # Add SIMD flags for the bytedelta filter and Intel (it seems that ARM64 does not need these)
         set_source_files_properties(
                 ${PROJECT_SOURCE_DIR}/plugins/filters/bytedelta/bytedelta.c
@@ -330,16 +324,10 @@ if(COMPILER_SUPPORT_AVX2)
         set_source_files_properties(
                 shuffle-avx2.c bitshuffle-avx2.c
                 PROPERTIES COMPILE_OPTIONS "/arch:AVX2")
-        set_property(
-                SOURCE shuffle.c
-                APPEND PROPERTY COMPILE_OPTIONS "/arch:AVX2")
     else()
         set_source_files_properties(
                 shuffle-avx2.c bitshuffle-avx2.c
                 PROPERTIES COMPILE_OPTIONS -mavx2)
-        set_property(
-                SOURCE shuffle.c
-                APPEND PROPERTY COMPILE_OPTIONS -mavx2)
     endif()
 
     # Define a symbol for the shuffle-dispatch implementation
@@ -354,16 +342,10 @@ if(COMPILER_SUPPORT_AVX512)
         set_source_files_properties(
                 bitshuffle-avx512.c
 		PROPERTIES COMPILE_OPTIONS "/arch:AVX512")
-        set_property(
-                SOURCE shuffle.c
-		APPEND PROPERTY COMPILE_OPTIONS "/arch:AVX512")
     else()
         set_source_files_properties(
                 bitshuffle-avx512.c
                 PROPERTIES COMPILE_OPTIONS "-mavx512f;-mavx512bw")
-        set_property(
-                SOURCE shuffle.c
-                APPEND PROPERTY COMPILE_OPTIONS "-mavx512f;-mavx512bw")
     endif()
 
     # Define a symbol for the shuffle-dispatch implementation
@@ -377,17 +359,11 @@ if(COMPILER_SUPPORT_NEON)
     set_source_files_properties(
             shuffle-neon.c bitshuffle-neon.c
             PROPERTIES COMPILE_OPTIONS "-flax-vector-conversions")
-    set_property(
-            SOURCE shuffle.c
-            APPEND PROPERTY COMPILE_OPTIONS "-flax-vector-conversions")
     if(CMAKE_SYSTEM_PROCESSOR STREQUAL armv7l)
         # Only armv7l needs special -mfpu=neon flag; aarch64 doesn't.
       set_source_files_properties(
             shuffle-neon.c bitshuffle-neon.c
             PROPERTIES COMPILE_OPTIONS "-mfpu=neon;-flax-vector-conversions")
-      set_property(
-            SOURCE shuffle.c
-            APPEND PROPERTY COMPILE_OPTIONS "-mfpu=neon;-flax-vector-conversions")
     endif()
     # Define a symbol for the shuffle-dispatch implementation
     # so it knows NEON is supported even though that file is
--- a/blosc/bitshuffle-altivec.c
+++ b/blosc/bitshuffle-altivec.c
@@ -24,6 +24,7 @@
 
 #include "bitshuffle-altivec.h"
 #include "bitshuffle-generic.h"
+#include <stdlib.h>
 
 /* Make sure ALTIVEC is available for the compilation target and compiler. */
 #if defined(__ALTIVEC__) && defined(__VSX__) && defined(_ARCH_PWR8)
@@ -33,7 +34,6 @@
 #include <altivec.h>
 
 #include <stdint.h>
-#include <stdlib.h>
 
 /* The next is useful for debugging purposes */
 #if 0
@@ -592,4 +592,23 @@ int64_t bshuf_untrans_bit_elem_altivec(const void* in, void* out, const size_t s
   return count;
 }
 
-#endif /* defined(__ALTIVEC__) */
+
+const bool is_bshuf_altivec = true;
+
+#else /* defined(__ALTIVEC__) && defined(__VSX__) && defined(_ARCH_PWR8) */
+
+const bool is_bshuf_altivec = false;
+
+int64_t
+bshuf_trans_bit_elem_altivec(const void* in, void* out, const size_t size,
+                             const size_t elem_size) {
+  abort();
+}
+
+int64_t
+bshuf_untrans_bit_elem_altivec(const void* in, void* out, const size_t size,
+                               const size_t elem_size) {
+  abort();
+}
+
+#endif /* defined(__ALTIVEC__) && defined(__VSX__) && defined(_ARCH_PWR8) */
--- a/blosc/bitshuffle-altivec.h
+++ b/blosc/bitshuffle-altivec.h
@@ -17,6 +17,12 @@
 
 #include <stddef.h>
 #include <stdint.h>
+#include <stdbool.h>
+
+/**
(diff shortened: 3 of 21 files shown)


FrancescAlted:
Thanks @mgorny for this.  However, your pull request will not use AVX2 in machines having it (you can easily check that by putting a `printf()` in https://github.com/Blosc/c-blosc2/blob/main/blosc/shuffle.c#L312).  This is because how the dynamic dispatching has been implemented inside blosc2; maybe there is a better way for implementing the dynamic dispatching, but again, your PR, as-is, is effectively disabling the use of AVX2.

Unfortunately AVX2 can greatly accelerate shuffle/unshuffle as well as bitshuffle/bitunshuffle, and it would be bad if users cannot leverage this capability anymore.

Also, @t20100 may have something to say here.

mgorny:
Oh, I'm sorry about that. I see the problem now, the `SHUFFLE_USE*` macros are the problem. Apparently they were added in dd57c035a6438cc3bf70efd852218c0ca4eddcab, but the commit message doesn't really give an example of the problem they solved. In fact, they explicitly contradict `CMakeLists.txt`:

https://github.com/Blosc/c-blosc2/blob/dc5d651c38dfb7b2527edeaa05bb493f050f5554/blosc/CMakeLists.txt#L345-L347

FWICS top-level `CMakeLists.txt` has a pretty specific list of compilers supporting AVX2, etc., so I'm surprised they would need to be double-checked like that. Of course, it would be much cleaner to perform a compile check rather than hardcode a list of compilers.

mgorny:
I've updated the pull request to fix the immediate issue. I can work on it further when I know what's the issue that prompted dd57c035a6438cc3bf70efd852218c0ca4eddcab.

t20100:
This commit (https://github.com/Blosc/c-blosc2/commit/dd57c035a6438cc3bf70efd852218c0ca4eddcab) was to add support of macos `universal2` builds (#431, #430). The issue in this case is that one compilation command builds both ARM and x86, so source files cannot be added conditionally to the build, nor macro defined per architecture.

On option proposed in https://github.com/Blosc/c-blosc/pull/347 (Same as #431 but for c-blosc1):

> An alternative could be to define `blosc_internal_*_avx2` and `blosc_internal_*_sse2` functions even when __AVX2__/__SSE2__ are not defined with a body calling e.g., abort().

With this `shuffle.c` can be built without SIMD options and all source files can be built on all architectures.

t20100:
As it is, I would expect `universal2` build to no longer work and any difference between what the CMakefile probes and SIMD macros like:
https://github.com/Blosc/c-blosc2/blob/dc5d651c38dfb7b2527edeaa05bb493f050f5554/blosc/bitshuffle-avx512.c#L24

to lead to issues.
I'll propose a change for this to clarify.

t20100:
Hope this helps to clarify: https://github.com/t20100/c-blosc2/commit/b243896db6262e4d99c5521e519a769c40ace773

It's only done for AVX2 and would need to be done for all SIMD implementations. It's maybe not necessary to do it for both shuffle and bitshuffle since the marco tests are the same.

It's basically adding a runtime test of the SIMD implementation availability and dummy functions when not available.
It still needs testing.

mgorny:
Should I add these to this PR or are you going to take it from here?

t20100:
> Should I add these to this PR or are you going to take it from here?

As you want.
I first wanted to get feedbacks before going further.

If it's OK, then you can either get my changes or I can make a PR to your branch to add it to this PR.

FrancescAlted:
@t20100 In general I like were you are headed (I just suggested to use `_AVX2` suffix).  It would be nice if both of you can contribute in this same PR (I don't if github makes this difficult though).

mgorny:
Ok, I'll update the PR in a few minutes.

mgorny:
> @t20100 In general I like were you are headed (I just suggested to use `_AVX2` suffix). It would be nice if both of you can contribute in this same PR (I don't if github makes this difficult though).

FWICS all the existing functions have `_AVX` suffix. Should I rename them?

mgorny:
(same goes for `SSE2` bits having `SSE` suffix)

FrancescAlted:
> > @t20100 In general I like were you are headed (I just suggested to use `_AVX2` suffix). It would be nice if both of you can contribute in this same PR (I don't if github makes this difficult though).
> 
> FWICS all the existing functions have `_AVX` suffix. Should I rename them?

Oh, I forgot about this.  Yeah, as they are internal functions, I think renaming is a good idea.

FrancescAlted:
> (same goes for `SSE2` bits having `SSE` suffix)

Let's do that here too.

mgorny:
One more quick question: I see that some header files (but not all) declare additional functions such as `bshuf_shuffle_bit_eightelem_altivec()` that aren't used elsewhere. Should I remove them from the headers while at it?

FrancescAlted:
> One more quick question: I see that some header files (but not all) declare additional functions such as `bshuf_shuffle_bit_eightelem_altivec()` that aren't used elsewhere. Should I remove them from the headers while at it?

Hmm, I'd say it is probably safe to remove those functions from the header, but I am not completely sure.  Perhaps @kif can shed some light here?

FrancescAlted:
> > One more quick question: I see that some header files (but not all) declare additional functions such as `bshuf_shuffle_bit_eightelem_altivec()` that aren't used elsewhere. Should I remove them from the headers while at it?
> 
> Hmm, I'd say it is probably safe to remove those functions from the header, but I am not completely sure. Perhaps @kif can shed some light here?

Thinking twice about this, let's focus in our problem at hand, and let the cleanup to happen in another PR.

mgorny:
Well, unfortunately this brings the second question: since these functions are technically present in the header, should I also stub them with `abort()`s? To be honest, that's some extra work that I don't really want to go through.

My rough idea for this PR would be to, in order:

1. Remove unneeded function decls from headers.
2. Rename functions for consistency.
3. Add checks and stubs as requested by @t20100.
4. And then, when everything's ready, remove the flags as this PR currently does.

FrancescAlted:
> Well, unfortunately this brings the second question: since these functions are technically present in the header, should I also stub them with `abort()`s? To be honest, that's some extra work that I don't really want to go through.
> 
> My rough idea for this PR would be to, in order:
> 
>     1. Remove unneeded function decls from headers.
> 
>     2. Rename functions for consistency.
> 
>     3. Add checks and stubs as requested by @t20100.
> 
>     4. And then, when everything's ready, remove the flags as this PR currently does.

I see. I'm +1 on your plan then.

mgorny:
Actually, I was wrong. Some of these functions are used cross-file, e.g. AVX512 uses AVX2 and SSE2 functions.

mgorny:
@t20100, actually, could you do the part with guards? I'm entirely unfamiliar with this code, and I'm getting too easily distracted for this.

t20100:
OK, I'll take care of the guards part.

t20100:
I made a PR on this branch with stubs when SIMD are not available and runtime availability checks: https://github.com/mgorny/c-blosc2/pull/1

What do you think?

mgorny:
Thanks a lot! I've merged it here.

FrancescAlted:
Hi.  I have tried this out, and it looks good to me.  If there is no opposition, I'll merge this soon.  Thanks @mgorny and @t20100 !

mgorny:
Thank you!

PR #626, opened 2024-07-14
load_lib: allow direct loading of library by name

boxerab:
This makes it possible to load plugins without relying on the presence of python


FrancescAlted (approved):
Other than a preferred max column size of 100, this looks good to me.

FrancescAlted:
Thanks @boxerab !

PR #631, opened 2024-09-09
Avoid rip-relative addressing for OSX x86_64

milesgranger:
G'day!

As part of [blosc2-rs](https://github.com/milesgranger/blosc2-rs) I find I need to add this caveat otherwise it fails on OSX x86_64 arch, but only when testing with static build of c-blosc2. 

Example failure run: https://github.com/milesgranger/blosc2-rs/actions/runs/10769365090/job/29860415491?pr=29#step:6:3029

> ld: invalid use of rip-relative addressing in '_get_shuffle_implementation' to '___cpu_model'

Maybe there is a proper way to fix this, or I'm doing something silly.

### DIFF
--- a/blosc/CMakeLists.txt
+++ b/blosc/CMakeLists.txt
@@ -37,6 +37,7 @@ if(BUILD_SHARED)
         $<INSTALL_INTERFACE:include>)
 endif()
 if(BUILD_STATIC)
+    add_compile_definitions(BUILD_STATIC)
     add_library(blosc2_static STATIC)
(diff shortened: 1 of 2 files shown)


milesgranger:
Looks like the Ubuntu GCC AARCH64 failure is unrelated, failing in [another workflow before](https://github.com/Blosc/c-blosc2/actions/runs/10548012229/job/29221396346) this PR.

FrancescAlted:
Thanks @milesgranger !

PR #634, opened 2024-10-14
Fix public include directories for blosc2 targets

dmikushin:
The `target_include_directories` interface is given in blosc2 subdirectory as `${CMAKE_CURRENT_SOURCE_DIR}`, which makes this directory automatically included when linking the blosc2 target in CMake. However, the blosc2 include directory is actually in the parent `../include` directory, so the correct way to include it could be `${CMAKE_CURRENT_SOURCE_DIR}/../include`. Moreover, `${CMAKE_CURRENT_SOURCE_DIR}` exposes internal headers instead of the public headers, so we may want to turn the original include directory into a private include.

### DIFF
--- a/blosc/CMakeLists.txt
+++ b/blosc/CMakeLists.txt
@@ -32,8 +32,9 @@ if(BUILD_SHARED)
         SOVERSION 4  # Change this when an ABI change happens
     )
     target_compile_definitions(blosc2_shared PUBLIC BLOSC_SHARED_LIBRARY)
+    target_include_directories(blosc2_shared PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
     target_include_directories(blosc2_shared PUBLIC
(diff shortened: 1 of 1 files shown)


dmikushin:
Hi @FrancescAlted , the CI on ARM is failing for an unrelated reason.

FrancescAlted:
Thanks @dmikushin .  Although I wonder if we can remove those lines entirely; after all, before this fix, they were buying nothing to configuration, right?

FrancescAlted:
LGTM. Thanks @dmikushin !

dmikushin:
Thanks!

> I wonder if we can remove those lines entirely

Ah, no, please don't remove them. I forgot to mention the usecase in my description. Actually, these lines absolutely make sense, when c-blosc2 is discovered by CMake via `find_package` AND when c-blosc2 integrates as a CMake `add_subdirectory` (usually as a `git add submodule`) into a larger software.

Issue #635, opened 2024-10-16
SEGV in blosc2_schunk_free()


EmilDohne:
**Describe the bug**
I noticed while upgrading from v2.11.2. -> 2.15.1. that I now receive a SEGV on calling blosc2_schunk_free(). 
Reverting back down to 2.14.0 fixes this problem. I was able to narrow it down to the 2.15.0 release.

The actual code calling blosc2_schunk_free() can be seen [here](https://github.com/EmilDohne/PhotoshopAPI/blob/feat-85-allow-for-swapping-of-image-data/PhotoshopAPI/src/Core/Struct/ImageChannel.h#L80)

As far as I'm aware I'm not doing anything that is unintended here but feel free to correct me

**Logs**
Below you can find some gh action runs across multiple OSs where these errors can be seen

[Valgrind 2.15.1](https://github.com/EmilDohne/PhotoshopAPI/actions/runs/11364823173/job/31611650862)
[Asan 2.15.1](https://github.com/EmilDohne/PhotoshopAPI/actions/runs/11364823168/job/31611650421)

[Valgrind 2.14.0](https://github.com/EmilDohne/PhotoshopAPI/actions/runs/11365462623/job/31613699739)
[Asan 2.14.0](https://github.com/EmilDohne/PhotoshopAPI/actions/runs/11365462630/job/31613700651) <-- There is some potentially unrelated issues in the GCC ubuntu build here

**System information:**
 - OS: Windows | Ubuntu | MacOS
 - Compiler GCC | Clang | MSVC
 - Version 2.15.0

EmilDohne:
I was able to narrow this issue down to the following lines in `blosc2_schunk_free()`:

```c
  if (schunk->storage != NULL) {
    blosc2_io_cb *io_cb = blosc2_get_io_cb(schunk->storage->io->id);
    if (io_cb != NULL) {
      int rc = io_cb->destroy(schunk->storage->io->params); // <----- io_cb->destroy is empty
      if (rc < 0) {
        BLOSC_TRACE_ERROR("Could not free the I/O resources.");
      }
    }
```

This seems to occur because in this function:

```c
blosc2_io_cb *blosc2_get_io_cb(uint8_t id) {
  for (uint64_t i = 0; i < g_nio; ++i) {
    if (g_ios[i].id == id) {
      return &g_ios[i];
    }
  }
  if (id == BLOSC2_IO_FILESYSTEM) {
[...]

PR #637, opened 2024-10-22
Fix SEGV in blosc2_schunk_free()

EmilDohne:
Apologies for only getting around to it now, this should address the recent regression in 2.15.0 :)

# Context:

#635 I noticed while upgrading to c-blosc2 2.15.0 that calling `blosc2_schunk_free()` without `blosc2_init()` being called first caused a SEGV as it was trying to access uninitialized memory

# Changes

- Added check to `blosc2_get_io_cb()` that returns `NULL` if the library was not initialized yet so we don't try to access a nullptr function in `blosc2_schunk_free()`
- Added test cases for "manual" compression and decompression using `blosc2_compress_ctx` and  `blosc2_decompress_ctx`. This does not in parallel but does trigger a test failure before my changes.

# Considerations

@FrancescAlted I noticed that `blosc2_destroy()` does not actually clean up the IO callbacks (note the order of tests, if it was reversed the tests would pass even without the fix). Is this intended?

### DIFF
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -4647,7 +4647,7 @@ int blosc2_register_io_cb(const blosc2_io_cb *io) {
 
 blosc2_io_cb *blosc2_get_io_cb(uint8_t id) {
   // If g_initlib is not set by blosc2_init() this function will try to read
-  // uninitialized memory
+  // uninitialized memory. We should therefore always return NULL in that case
   if (!g_initlib) {
     return NULL;
   }
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -4646,6 +4646,11 @@ int blosc2_register_io_cb(const blosc2_io_cb *io) {
 }
 
 blosc2_io_cb *blosc2_get_io_cb(uint8_t id) {
+  // If g_initlib is not set by blosc2_init() this function will try to read
+  // uninitialized memory
+  if (!g_initlib) {
+    return NULL;
+  }
   for (uint64_t i = 0; i < g_nio; ++i) {
     if (g_ios[i].id == id) {
       return &g_ios[i];
--- a/tests/test_schunk.c
+++ b/tests/test_schunk.c
@@ -168,7 +168,7 @@ static char* test_schunk(void) {
 static char* test_schunk_no_init(void)
 {
   // Test that calling blosc2_schunk without a call to blosc2_init() is valid and works as intended
-  // as this is the recommended route for multithreaded 
+  // as this is the recommended route for multithreaded compression/decompression
   static int32_t data[CHUNKSIZE];
   static int32_t data_dest[CHUNKSIZE];
   int32_t isize = CHUNKSIZE * sizeof(int32_t);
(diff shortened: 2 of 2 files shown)


EmilDohne:
I think the error is because clang didnt like the missing return type which I missed. I've also gone ahead and address some of the unused variable warnings. Hopefully it passes CI now!

FrancescAlted:
> I think the error is because clang didnt like the missing return type which I missed. I've also gone ahead and address some of the unused variable warnings. Hopefully it passes CI now!

Yes, your fix did it, at least on my mac laptop.  I am currently traveling, but I'll come back to you later.

Regarding `blosc2_destroy()` not cleaning up I/O calls is probably an oversight. Can you fix this too?

EmilDohne:
> Regarding `blosc2_destroy()` not cleaning up I/O calls is probably an oversight. Can you fix this too?

I'll have a deeper look into this and see if it needs to be cleaned up. Enjoy your travels!

EmilDohne:
Looks like this was also the reason for the segfaults in the aarch tests :)

EmilDohne:
Ok so I had a look into the cleanup stuff and at least from a memory standpoint there is no concerns. One could, if wanted, set the se back to be default initialized but I wouldn't personally do that

```c
// In blosc2_init()

BLOSC2_IO_CB_DEFAULTS.id = BLOSC2_IO_FILESYSTEM;
BLOSC2_IO_CB_DEFAULTS.name = "filesystem";
BLOSC2_IO_CB_DEFAULTS.is_allocation_necessary = true;
BLOSC2_IO_CB_DEFAULTS.open = (blosc2_open_cb) blosc2_stdio_open;
BLOSC2_IO_CB_DEFAULTS.close = (blosc2_close_cb) blosc2_stdio_close;
BLOSC2_IO_CB_DEFAULTS.size = (blosc2_size_cb) blosc2_stdio_size;
BLOSC2_IO_CB_DEFAULTS.write = (blosc2_write_cb) blosc2_stdio_write;
BLOSC2_IO_CB_DEFAULTS.read = (blosc2_read_cb) blosc2_stdio_read;
BLOSC2_IO_CB_DEFAULTS.truncate = (blosc2_truncate_cb) blosc2_stdio_truncate;
BLOSC2_IO_CB_DEFAULTS.destroy = (blosc2_destroy_cb) blosc2_stdio_destroy;

BLOSC2_IO_CB_MMAP.id = BLOSC2_IO_FILESYSTEM_MMAP;
BLOSC2_IO_CB_MMAP.name = "filesystem_mmap";
BLOSC2_IO_CB_MMAP.is_allocation_necessary = false;
BLOSC2_IO_CB_MMAP.open = (blosc2_open_cb) blosc2_stdio_mmap_open;
BLOSC2_IO_CB_MMAP.close = (blosc2_close_cb) blosc2_stdio_mmap_close;
BLOSC2_IO_CB_MMAP.read = (blosc2_read_cb) blosc2_stdio_mmap_read;
BLOSC2_IO_CB_MMAP.size = (blosc2_size_cb) blosc2_stdio_mmap_size;
BLOSC2_IO_CB_MMAP.write = (blosc2_write_cb) blosc2_stdio_mmap_write;
BLOSC2_IO_CB_MMAP.truncate = (blosc2_truncate_cb) blosc2_stdio_mmap_truncate;
BLOSC2_IO_CB_MMAP.destroy = (blosc2_destroy_cb) blosc2_stdio_mmap_destroy;
```

From my side there is nothing else to add so feel free to merge :)

FrancescAlted:
LGTM. Thanks @EmilDohne !

Issue #636, opened 2024-10-16
UB in shuffle_avx2 and unshuffle_avx2


EmilDohne:
**Describe the bug**

These two typedefs suggest that both `_src` and `_dest` should be `const uint8_t*`.

```c
typedef void(* shuffle_func)(const int32_t, const int32_t, const uint8_t*, const uint8_t*);
typedef void(* unshuffle_func)(const int32_t, const int32_t, const uint8_t*, const uint8_t*);
```
However, the avx2 implementation does not take a `const uint8_t*` but instead a `uint8_t*`. On further inspection this seems to hold true for other implementations. To my knowledge, all calls to shuffle actually call with a non-const pointer so I think it's likely fine to make these values non-const

**Logs**
Below you can find a `fsanitize=undefined` run that points this out:

https://github.com/EmilDohne/PhotoshopAPI/actions/runs/11365462630/job/31613700651

FrancescAlted:
This sounds reasonable to me.  Would you mind providing a PR?

EmilDohne:
I can definitely have a look at providing one!

EmilDohne:
Fixed by #638

PR #638, opened 2024-10-22
fix UB in shuffle and unshuffle by marking _dst non-const

EmilDohne:
# Context:
https://github.com/Blosc/c-blosc2/issues/636 I noticed while running my library that relies on blosc2 under `fsanitize=undefined` that the shuffle implementations did not all have `_dest` marked as const while the function calling it did. Upon further investigation the function calling shuffle and unshuffle did not have any of these marked as const so there was no deeper impact but I wanted to address it anyways.

# Changes:

- make shuffle and unshuffle funcs take a non-const `uint8_t*` for `_dest`
- Checked that none of the shuffle implementations also had their `_src` marked as non-const.

### DIFF
--- a/blosc/shuffle.c
+++ b/blosc/shuffle.c
@@ -59,8 +59,8 @@
 
 
 /*  Define function pointer types for shuffle/unshuffle routines. */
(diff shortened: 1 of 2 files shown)


FrancescAlted:
LGTM. Thanks @EmilDohne !

PR #648, opened 2025-02-04
12-byte unshuffle for avx2

froody:


FrancescAlted:
That is very cool.  Yes, shuffling for smaller data sizes that are powers of 2 is very interesting, most specially for records with fields of atomic types.  Thanks @froody !

PR #650, opened 2025-02-05
Add 12-byte sse2 unshuffle

froody:


FrancescAlted:
Cool! Thanks @froody !

PR #655, opened 2025-02-26
Fixed issue #633 Handle leak in parallel (nThread > 1) zstd decompres…

### LINKED ISSUE #633: Handle leak in parallel (nThread > 1) zstd decompression
**Describe the bug**

Resources leak has been observed in blosc2_decompress_ctx  for zstd with nthreads > 1.

It seems that thread resources are not properly released when zstd is decompressed in parallel.
- if nThreads == 1, everything is fine.
- if nThreads is > 1 each decompression increases the amount of used handles in the System.

Though the leak of a single decompression is quite small and neglectable, it leads inevitable to unacceptable behavior in long running application with very frequent blosc2_decompress_ctx calls.

Bug was initially found in blosc2 version 2.14.4 and is still reproducible in blosc2 version 2.15.1.

**To Reproduce**
Use blosc2_decompress_ctx with nthreads > 1 and zstd. Decompress lots of objects and observe handle/thread consumption.

**Expected behavior**
Threads are released.

**Logs**

**System information:**
* OS: Windows 10
* Compiler: Microsoft (R) C/C++ Optimizing Compiler Version
* Version: 19.38.33130
* Blosc version: 2.15.1

**Additional context**

jocbeh:
This fix resolves #633.

The bug only affects windows platform in case de-compression is configured to be processed in parallel. Worker threads are not closed due to missing CloseHandle in file win32/pthread.c

Fix has been tested successfully in our windows application.

Note: This fix is inspired by the following commit: https://git.ipfire.org/?p=thirdparty/git.git;a=commit;f=compat/win32/pthread.c;h=238a9dfe86d18214bc0b8be079feee01a2fdc407

### DIFF
--- a/blosc/win32/pthread.c
+++ b/blosc/win32/pthread.c
@@ -77,8 +77,10 @@ int win32_pthread_join(pthread_t *thread, void **value_ptr)
 		case WAIT_OBJECT_0:
 			if (value_ptr)
 				*value_ptr = thread->arg;
+			CloseHandle(thread->handle);
 			return 0;
 		case WAIT_ABANDONED:
+			CloseHandle(thread->handle);
 			return EINVAL;
 		default:
 			return GetLastError();


FrancescAlted:
Good catch.  Thanks @jocbeh !

Issue #656, opened 2025-02-28
heap-buffer-overflow in compress_chunk_fuzzer


lmarch2:
# Description
A heap-buffer-overflow was detected in `compress_chunk_fuzzer `with oss-fuzz on commit `16450518afddcb3139de627157208e49bfef6987 `. It occurred in function `_sw32` that causes an overflow due to an unexpected read from an allocated buffer.
# Details
```
==8==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x504000003030 at pc 0x5ea20ff546bb bp 0x7ffdf88637a0 sp 0x7ffdf8863798
WRITE of size 4 at 0x504000003030 thread T0
SCARINESS: 36 (4-byte-write-heap-buffer-overflow)
    #0 0x5ea20ff546ba in _sw32 /src/c-blosc2/blosc/blosc-private.h:119:23
    #1 0x5ea20ff546ba in serial_blosc /src/c-blosc2/blosc/blosc2.c:1958:7
    #2 0x5ea20ff546ba in do_job /src/c-blosc2/blosc/blosc2.c:2156:15
    #3 0x5ea20ff3e8e9 in blosc_compress_context /src/c-blosc2/blosc/blosc2.c:2480:15
    #4 0x5ea20ff405e7 in blosc2_compress /src/c-blosc2/blosc/blosc2.c:2906:12
    #5 0x5ea20ff34312 in LLVMFuzzerTestOneInput /src/c-blosc2/tests/fuzz/fuzz_compress_chunk.c:44:15
    #6 0x5ea20fde8a90 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:614:13
    #7 0x5ea20fde82b5 in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:516:7
    #8 0x5ea20fde9a95 in fuzzer::Fuzzer::MutateAndTestOne() /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:760:19
[...]

PR #658, opened 2025-03-23
Fix various UB as well as compiler warnings 

EmilDohne:
This MR aims to touch on a couple of instances of undefined behaviour in the c-blosc2 library as well as various warnings emitted by compilers 

- Fix uninitialized memory access in newly added `unshuffle12_sse2` and `unshuffle12_avx2` functions
- Fix unaligned access in `_sw32` and `sw32_` I've had a look at where this function is used and it looks like it won't affect performance in any way. I wasnt able to figure out why the access was unaligned but memcpy is safe on unaligned operations
- Fix `DWORD` being printed as `%s` in `sprintf` call
- Fix warning on unused variable (since this variable was only being used in the linux branch)
- `splitmode` variable was uninitialized if goto was triggered

Let me know if you have comments on any of these :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ cmake-build-*
 /blosc/config.h
 /doc/doxygen/xml/
 /doc/xml
+.vs/
--- a/blosc/blosc-private.h
+++ b/blosc/blosc-private.h
@@ -280,7 +280,7 @@ static inline void* load_lib(char *plugin_name, char *libpath) {
         return loaded_lib;
     } else {
 #if defined(_WIN32)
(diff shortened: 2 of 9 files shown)


FrancescAlted:
Thanks @EmilDohne !  From your fixes, the one for  `_sw32` and `sw32_` intrigued me as well; but as this is not critical for performance, I am fine with it.

EmilDohne:
Thanks for merging! I wonder if that is not related to this, although in my use cases I never have a block that is just `BLOSC2_MAX_OVERHEAD` https://github.com/Blosc/c-blosc2/issues/656

PR #660, opened 2025-05-11
flush modified pages only in write mode for mmap files

JanSellner:
Fixes issue [https://github.com/Blosc/python-blosc2/issues/389](https://github.com/Blosc/python-blosc2/issues/389)

Flushing modified pages only makes sense for write modes (w+ and r+). This slipped though because there was no error checking of `blosc2_schunk_free` which I now also added.


FrancescAlted:
LGTM.  Thanks @JanSellner !

PR #662, opened 2025-06-10
Fix concatenate

lshaw8317:
Add arange capability to test and use get_slice on chunks instead of decompressing chunks into buffer before loading into target array (thereby avoiding reordering of elements according to block shape).


FrancescAlted:
Cool.  However, this PR ended having the whole set of files in repo.  This is likely due to your CLion config: newlines seem to be set to CR/LF (the standard in Windows), whereas the UNIX standard is using just a LF.  You can change this by going to the main menu, go to File | File Properties | Line Separators, and then select a LF style from the list.  Better close this and open a new one.

PR #664, opened 2025-06-13
Added expand_dims function and tests

lshaw8317:
In order to implement stack function in later version, added a C-level ``expand_dims`` function (inverse of ``squeeze``).


lshaw8317:
This implementation will not work for contiguous arrays since by enlarging the shape one enlarges the metadata.

PR #665, opened 2025-06-17
First commit to include expand_dims

lshaw8317:
Begin to add functionality of views to b2nd arrays. This PR only effectively implements views which have an extra dummy dimension, via the function expand_dims. However, by introducing schunks with a view flag which can be created by the (private) make_view function, and modifying the destructor blosc2_schunk_free, it should be possible to extend the code to allow for e.g. slices as views.
Copying views (to contiguous and/or persistent files) also seems to be supported by the code base with the small modifications of this PR.


FrancescAlted (commented):
Good job!  I have some comments.  Also, remember to add new `b2nd_expand_dims` to "doc/reference/b2nd.rst" (see e.g. 37ab2b3d8bc85d11d2b055af947648c9810dfea7).

PR #670, opened 2025-07-03
Eagerly register io_cb's to avoid race

froody:
I am seeing a crash on darwin/arm64 when calling blosc2 from multiple threads concurrently. I serialize the call to `blosc2_init` but this doesn't prevent a race when multiple threads call _blosc2_register_io_cb simultaneously. See stacktrace of crash below.
```
Thread 13 Crashed:
0   libsystem_platform.dylib 0x1935c31b0 _platform_strcmp$VARIANT$Base + 64
1   pointcloud_sampler       0x10543b218 _blosc2_register_io_cb + 132
2   pointcloud_sampler       0x10543b3ec blosc2_get_io_cb + 188
3   pointcloud_sampler       0x105446ad0 frame_from_schunk + 936
4   pointcloud_sampler       0x10544141c blosc2_schunk_new + 1000
```

### DIFF
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -3818,6 +3818,8 @@ blosc2_io *blosc2_io_global = NULL;
 blosc2_io_cb BLOSC2_IO_CB_DEFAULTS;
 blosc2_io_cb BLOSC2_IO_CB_MMAP;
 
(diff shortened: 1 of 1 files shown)


FrancescAlted:
I am fine with this. Maybe @EmilDohne has something to say here?

EmilDohne:
LGTM too

PR #674, opened 2025-08-04
Include pthread.h instead of pthread.c

barakugav:


FrancescAlted:
PR is not passing tests for MSVC.

barakugav:
> This ended being a bit more intrusive than I would have preferred. Thing is, IIRC `blosc/win32/threading.c` is vendored from git sources, and improvements/fixes from there would be more difficult to port after this PR. It would be nice if a not so-intrusive solution can be devised.

Oh I didnt know it was vendored. The only change it in is the prefix of `blosc2_` to the function names, is that a blocker?
Its hard for me to come up with another solution if we dont want the compiler to export these symbols, they can not be in the .c file...
Tell me what you think

FrancescAlted:
Your last commit looks good to me.  Any reason why you renamed blosc/win32/pthread.c → blosc/win32/threading.c ?  Although, after inspection of current [win32/pthread.c](https://github.com/git/git/blob/2c2ba49d55ff26c1082b8137b1ec5eeccb4337d1/compat/win32/pthread.c) in git sources, they changed the code a lot since we have vendored the file, so I think the renaming is good as well.

Is this ready for merging?

barakugav:
pthread is specific to UNIX so I thought `threading.h` is a more appropriate name for the unified header, and win32/threading.c is an implementation of it so using the same name as the header felt most suitable.

And yes, this PR is ready for merge :)

FrancescAlted:
Thanks again @barakugav !

PR #677, opened 2025-08-08
Add unified `Blosc2::blosc2` CMake target.

teo-tsirpanis:
This PR adds a `Blosc2::blosc2` CMake target that is always available regardless of the linkage of the built library. If both a static and shared library were built, the target will use the linkage based on the value of the `BUILD_SHARED_LIBS` variable.

Similar unified targets are also available in [lz4](https://github.com/lz4/lz4/blob/cacca37747572717ceb1f156eb9840644205ca4f/build/cmake/CMakeLists.txt#L160-L167) and [zstd](https://github.com/facebook/zstd/blob/e128976193546dceb24249206a02ff8f444f7120/build/cmake/lib/CMakeLists.txt#L156-L185).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -445,6 +445,28 @@ if(BUILD_PLUGINS)
     add_subdirectory(plugins)
 endif()
 
+# Determine which linkage the unified target will use
+if(BUILD_SHARED AND NOT BUILD_STATIC)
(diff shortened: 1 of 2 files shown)


FrancescAlted:
Thanks @teo-tsirpanis !  Should this affect to upstream projects (like python-blosc2?), or they would compile without changes?

PR #678, opened 2025-08-08
Fix detection of AVX512 in MSVC.

teo-tsirpanis:


FrancescAlted:
Thanks @teo-tsirpanis !  Have you tested it on a machine with AVX512?  Out of curiosity, do you see any speed-ups?

PR #681, opened 2025-08-13
Add public APIs for raw shuffle operations.

### LINKED ISSUE #680: Feature request: Add public APIs for raw shuffling.
[TileDB](https://github.com/TileDB-Inc/TileDB) supports (bit)shuffle among its data compression [filters](https://cloud.tiledb.com/academy/structure/arrays/foundation/key-concepts/storage/tile-filters/index.html). For their implementation it's been using a vendored copy of blosc and https://github.com/kiyo-masui/bitshuffle, for shuffle and bitshuffle respectively.

We would like to remove the vendored sources and add Blosc2 as a proper dependency, but right now there are no public APIs for just the (bit)(un)shuffle operations. In order to resolve that, I am proposing to add the following functions to the library's public API:

```c
int blosc2_shuffle(const int32_t typesize, const int32_t blocksize, const uint8_t* src, uint8_t* dest);

int blosc2_unshuffle(const int32_t typesize, const int32_t blocksize, const uint8_t* src, uint8_t* dest);

int blosc2_bitshuffle(const int32_t typesize, const int32_t blocksize, const uint8_t* src, uint8_t* dest);

int blosc2_bitunshuffle(const int32_t typesize, const int32_t blocksize, const uint8_t* src, uint8_t* dest);
```

teo-tsirpanis:
Fixes #680.

### DIFF
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -984,10 +984,10 @@ uint8_t* pipeline_forward(struct thread_context* thread_context, const int32_t b
     if (filters[i] <= BLOSC2_DEFINED_FILTERS_STOP) {
       switch (filters[i]) {
         case BLOSC_SHUFFLE:
-          shuffle(typesize, bsize, _src, _dest);
+          blosc2_shuffle(typesize, bsize, _src, _dest);
           break;
         case BLOSC_BITSHUFFLE:
-          if (bitshuffle(typesize, bsize, _src, _dest) < 0) {
+          if (blosc2_bitshuffle(typesize, bsize, _src, _dest) < 0) {
             return NULL;
           }
           break;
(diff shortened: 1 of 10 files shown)


FrancescAlted:
Ok.  I see that you are adding new .cdata files, which is great.  Can you rename them to add the current version (currently 2.20) as per other files in the compat/ dir?  Bonus points if you add [quick instructions to check for forward compatibility before a release](https://github.com/Blosc/c-blosc2/blob/1649e4425983583eb266266190fa1ac4e980cea5/RELEASING.rst?plain=1#L28-L53).

FrancescAlted:
Also, I am still missing *test units* that actually exercices the new `blosc2_shuffle` and friends calls.  This is important to avoid unadverted or subtle API changes in the future.  Adding example in the examples/ directory is a good practice too.

teo-tsirpanis:
Added roundtrip tests for the public APIs, as well as the version number to the compat files. I don't have the bandwidth to write examples right now.

FrancescAlted:
LGTM.  An example can be added later.  Thanks @teo-tsirpanis !

PR #686, opened 2025-08-16
Use newer Win32 synchronization primitives.

teo-tsirpanis:
This PR updates `threading.h` on Windows to take advantage of newer synchronization primitives, introduced in Windows Vista.

* `blosc2_pthread_mutex_t` was changed from an alias to `CRITICAL_SECTION`, to an alias to [`SRWLOCK`](https://learn.microsoft.com/en-us/windows/win32/Sync/slim-reader-writer--srw--locks).
  * One difference from critical sections is that SRW locks are not recursive, which is also the case with pthread mutexes by default, and we don't currently use recursive mutexes anywhere in the codebase.
* `blosc2_pthread_cond_t` was updated to be an alias to [`CONDITION_VARIABLE`](https://learn.microsoft.com/en-us/windows/win32/sync/condition-variables). The custom implementation of condition variables was removed.

There's also a [synchronization barrier API](https://learn.microsoft.com/en-us/windows/win32/sync/synchronization-barriers) that we could make use of, but the transition will be slightly less trivial.


teo-tsirpanis:
Closing for now due to deadlocks in `test_b2nd_append` that I also observed locally, and occur with either `CRITICAL_SECTION` or `SRWLOCK`.

PR #387, opened 2022-04-05
Add support for zlib 2.0.6

FrancescAlted:

Issue #687, opened 2025-08-18
cb593a2c8667fbf72475d4c9953668c63f0df346 broke support for building against system libraries


mgorny:
**Describe the bug**
The changes done in cb593a2c8667fbf72475d4c9953668c63f0df346 (with followup fixes) make it possible to build against external lz4, zlib-ng and zstd **only** if they are built with CMake. However:

- [lz4 supports four different build systems](https://github.com/lz4/lz4/tree/dev/build)
- [zlib-ng supports building either with CMake or configure](https://github.com/zlib-ng/zlib-ng?tab=readme-ov-file#build)
- [zstd supports three different build systems](https://github.com/facebook/zstd/tree/dev/build)

This change at least broke compatibility with Fedora (that builds lz4 with Meson and zstd with Make) and Gentoo (that builds lz4 and zstd with Meson). Also, given that Meson is a mode modern build system that CMake, there is no reason for these distributions to use CMake, especially that it creates installations that aren't compatible with other builds of these packages.

**To Reproduce**

(e.g. using a Fedora or Gentoo install)

```
cmake ${srcdir} -G Ninja -DBUILD_SHARED_LIBS=ON -DBUILD_STATIC=OFF -DBUILD_TESTS=ON -DPREFER_EXTERNAL_{LZ4,ZLIB,ZSTD}=ON
```

**System information:**
 - OS: Gentoo Linux amd64
 - Compiler: gcc 15.2.0
 - Version: 2.1.0

**Additional context**

PR #690, opened 2025-08-18
Add fallbacks if no CMake config was found et al.

teo-tsirpanis:
Fixes #387.

This PR adds back the find modules removed in #682, but uses them as a fallback if the corresponding config file was not found. We hardcode the availability of each dependency's config file, to our own config file, and call `find_dependency` based on that.

Validated by building the vcpkg port with this PR applied, as well as by building a modified port that used `zlib-ng`, the latter of which uncovered a bug.


FrancescAlted:
LGTM. Thanks @teo-tsirpanis and @mgorny !

FrancescAlted:
> Fixes #387.

You probably wanted to say #687

PR #692, opened 2025-08-21
Allocate strides array on stack if ndim <= B2ND_MAX_DIM in unidim_to_multidim

barakugav:


blosc/blosc2.c
```diff
@@ -4706,7 +4717,15 @@ void blosc2_unidim_to_multidim(uint8_t ndim, int64_t *shape, int64_t i, int64_t
   if (ndim == 0) {
     return;
   }
-  int64_t *strides = malloc(ndim * sizeof(int64_t));
+
+  int64_t strides_buf[B2ND_MAX_DIM];
+  int64_t *strides;
+  if (BLOSC_LIKELY(ndim <= B2ND_MAX_DIM)) {
+    strides = strides_buf;
+  } else {
```
FrancescAlted, line 4725: I think you can make this simpler and assume that `ndim <= B2ND_MAX_DIM` is always fullfilled.

FrancescAlted: It can't be, because `ndim > B2ND_MAX_DIM` cannot possibly work.

PR #697, opened 2025-09-17
Fix compatibility with glibc v2.42

avalentino:
Before checking a possible error of the strtol() function the 'errno' have to be reset to zero.
Apparently there are situations in which 'errno' can be legally set to a non-zaro value and this interferes with the error checking of strtol().

Without the patch, the unit test suite stalls when it is executed in an environment with glibc v2.42 and linux kernel < v6.13.

Patch kindly provided by Aurelien Jarno.
Original bub report:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1115531.


FrancescAlted:
Thanks @avalentino !

PR #727, opened 2026-04-03
Harden lazy chunk parsing against integer overflow and malformed headers

metsw24-max:
Fixes memory-safety issues in frame_get_lazychunk due to insufficient validation of header-derived values
Addresses integer overflow risks in:
`nblocks * sizeof(int32_t)`
`trailer_offset + trailer_len`
Adds strict validation for `nblocks` (rejects 0 and values > INT_MAX)
Introduces overflow checks for all critical size calculations (multiplication and addition)
Adds bounds validation for offsets and computed sizes to prevent out-of-bounds access
Validates all memory allocations and ensures safe handling of failures
Ensures proper cleanup of allocated memory on all error paths
Updates/adds regression test in `tests/test_vlblocks.c` using `UINT32_MAX` to simulate malformed input
Verifies that invalid input is safely rejected without unsafe behavior
Improves robustness against corrupted or malicious inputs with no impact on valid cases

### DIFF
--- a/blosc/frame.c
+++ b/blosc/frame.c
@@ -23,6 +23,7 @@
 #endif  /* _WIN32 */
 
 #include <inttypes.h>
+#include <limits.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdint.h>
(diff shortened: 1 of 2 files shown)


FrancescAlted:
LGTM. Thanks @metsw24-max !

PR #730, opened 2026-04-10
 harden metadata deserialization with bounds and length validation

metsw24-max:
This PR hardens `b2nd_deserialize_meta` against malformed or truncated metadata by introducing strict bounds checking and validation of length fields before reading from the input buffer.

The parser previously assumed well-formed input and could advance pointers and read fields without verifying that sufficient bytes remained. This could lead to out-of-bounds reads and unsafe memory operations when handling corrupted or adversarial inputs.

-Added systematic bounds checks before reading each metadata field
-Validated `dtype_len` to ensure it is non-negative and within remaining buffer limits
-Added early checks for null or invalid input buffers
-Ensured safe failure on malformed metadata instead of continuing parsing

### DIFF
--- a/blosc/b2nd.c
+++ b/blosc/b2nd.c
@@ -188,6 +188,10 @@ int b2nd_deserialize_meta(const uint8_t *smeta, int32_t smeta_len, int8_t *ndim,
     // dtype info is here
     B2ND_REQUIRE_META_NBYTES(1 + 1 + sizeof(int32_t));
     *dtype_format = (int8_t) *(pmeta++);
+    if (*pmeta != 0xdb) {
+      BLOSC_TRACE_ERROR("Malformed b2nd metalayer: invalid dtype MsgPack marker");
+      return BLOSC2_ERROR_FAILURE;
+    }
     pmeta += 1;
     int32_t dtype_len;
     swap_store(&dtype_len, pmeta, sizeof(int32_t));
@@ -665,7 +669,7 @@ int b2nd_get_slice_nchunks(const b2nd_array_t *array, const int64_t *start, cons
 
   int8_t ndim = array->ndim;
 
-  if (array->nitems == 0){
+  if (array->nitems == 0) {
     *chunks_idx = NULL;
     return 0;
   }
--- a/blosc/b2nd.c
+++ b/blosc/b2nd.c
@@ -102,6 +102,12 @@ int b2nd_deserialize_meta(const uint8_t *smeta, int32_t smeta_len, int8_t *ndim,
   BLOSC_ERROR_NULL(shape, BLOSC2_ERROR_NULL_POINTER);
   BLOSC_ERROR_NULL(chunkshape, BLOSC2_ERROR_NULL_POINTER);
   BLOSC_ERROR_NULL(blockshape, BLOSC2_ERROR_NULL_POINTER);
+  if (dtype != NULL) {
+    *dtype = NULL;
+  }
+  if (dtype_format != NULL) {
+    *dtype_format = 0;
+  }
   if (smeta_len <= 0) {
     BLOSC_TRACE_ERROR("Malformed b2nd metalayer: empty metadata");
     return BLOSC2_ERROR_FAILURE;
@@ -178,7 +184,6 @@ int b2nd_deserialize_meta(const uint8_t *smeta, int32_t smeta_len, int8_t *ndim,
   if (dtype_format == NULL || dtype == NULL) {
     return (int32_t)(pmeta - smeta);
   }
-  *dtype = NULL;
   if (pmeta - smeta < smeta_len) {
     // dtype info is here
     B2ND_REQUIRE_META_NBYTES(1 + 1 + sizeof(int32_t));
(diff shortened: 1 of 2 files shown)


metsw24-max:
should i apply these changes?

FrancescAlted:
I am not asking to apply these changes blindly.  I just want your opinion on these, and if applicable, yes, let's use them.

metsw24-max:
I reviewed the suggestions carefully and went ahead with the two changes that are directly related to parsing safety and correctness:

>smeta_end = smeta + smeta_len can invoke undefined behavior if smeta_len is large enough to overflow the pointer range (the API accepts caller-provided pointers/lengths). Consider avoiding end-pointer computation and instead track remaining bytes using integer arithmetic (e.g., size_t remaining = (size_t)smeta_len - (size_t)(pmeta - smeta)), updating B2ND_REQUIRE_META_NBYTES accordingly.


- Reworked B2ND_REQUIRE_META_NBYTES to use a consumed/remaining approach instead of end-pointer arithmetic, to avoid potential undefined behavior with malformed metadata lengths.

> `dtype_len` is read as a 32-bit value from the buffer (`sizeof(int32_t)`), but it's currently stored in an `int`. On platforms where `int` is not 32-bit, this can lead to incorrect reads or memory corruption. Use `int32_t dtype_len` to match the serialized format and the `swap_store` size.
- Updated dtype_len from int to int32_t to match the serialized 32-bit field and avoid any platform-dependent issues.

>To ensure safe failure semantics, initialize output parameters as early as possible. Right now *dtype = NULL is only set after successfully parsing up through blockshape; if parsing fails earlier, callers that don't pre-initialize dtype (e.g., bench/b2nd/bench_zfp_getitem.c) could end up freeing an uninitialized pointer. Suggest setting *dtype = NULL (and optionally *dtype_format = 0) immediately after validating that dtype/dtype_format are non-NULL.

For the second suggestion (initializing dtype / dtype_format earlier), I considered it more of an API-hardening improvement rather than a parsing safety issue, so I kept this PR focused on the core robustness fixes.

i will do the changes on 2nd suggestion also on your preference

FrancescAlted:
Yes, I do think that doing the changes on 2nd suggestion is interesting indeed. Can you apply that too?

metsw24-max:
I reviewed that suggestion and applied the failure-semantics part without changing API behavior.

I did not make `dtype`/`dtype_format` mandatory because `b2nd_deserialize_meta` currently supports optional dtype parsing and several in-tree callers pass `NULL, NULL` intentionally.

Current approach initializes outputs early when non-NULL while preserving the existing optional-argument contract.

FrancescAlted:
Ok, your changes look good to me. Now, Copilot came with a couple of new suggestions. Thoughts?

metsw24-max:
I checked the new Copilot suggestions 

>The deserializer skips over the MsgPack array header and version byte without validating that they match the expected format (e.g., fixarray with 7 entries and B2ND_METALAYER_VERSION). With only bounds checks, malformed metadata with unexpected type tags can still be parsed into arbitrary ndim/shape values, which can lead to downstream inconsistencies. Consider validating these marker bytes (and the fixarray lengths for shape/chunkshape/blockshape) and failing fast when they don’t match the expected encoding.

The header validation looks like a good improvement right now we’re mostly doing bounds checks, so verifying the MsgPack markers would make things more robust.

>In the dtype section, the code does pmeta += 1 to skip the MsgPack string header, but it never checks that the byte is actually 0xdb (as emitted by b2nd_serialize_meta). If the metadata is malformed but long enough, dtype_len will be read from the wrong offset and could yield nonsensical values. Suggest validating the expected type marker(s) here (and optionally that *dtype_format is within supported values) before reading dtype_len.

Same with the dtype part  since we always write a str32 marker on the serialize side, it makes sense to validate it before skipping. That should help avoid any parsing issues with malformed input.

>Minor style: there’s still no space before the opening brace (if (array->nitems == 0){). To match the surrounding style in this file, consider if (array->nitems == 0) {.

The brace comment feels like just a style thing so I don’t think it’s important for this PR.

Only thing I would be a bit careful about is enforcing a strict array size (like fixarray(7)), since there seems to be some backward compatibility with older metadata. Maybe better to validate structure without being too strict there.

Overall the first two suggestions look like good hardening improvements

should i implement?

FrancescAlted:
Yes, if you think that the suggestions are good, +1 on implementing them.

>> Minor style: there’s still no space before the opening brace (if (array->nitems == 0){). To match the surrounding style in this file, consider if (array->nitems == 0) {.

> The brace comment feels like just a style thing so I don’t think it’s important for this PR.

But it is important for keeping the code base nicely formatted, so +1 on correcting this too.

> Only thing I would be a bit careful about is enforcing a strict array size (like fixarray(7)), since there seems to be some backward compatibility with older metadata. Maybe better to validate structure without being too strict there.

+1

metsw24-max:
i implemented 2nd and 3rd suggested changes this ran all checks successfully

and also i implemented the first suggestion:
>The deserializer skips over the MsgPack array header and version byte without validating that they match the expected format (e.g., fixarray with 7 entries and B2ND_METALAYER_VERSION). With only bounds checks, malformed metadata with unexpected type tags can still be parsed into arbitrary ndim/shape values, which can lead to downstream inconsistencies. Consider validating these marker bytes (and the fixarray lengths for shape/chunkshape/blockshape) and failing fast when they don’t match the expected encoding.

I first tried enforcing the expected array size, which caused test failures. I then relaxed it to only validate the array marker without enforcing a strict size, but that still led to failures in existing tests

so i kept it to 2nd and 3rd suggestion

FrancescAlted (approved):
LGTM

FrancescAlted:
Thanks @metsw24-max !

PR #731, opened 2026-04-13
Fix out-of-bounds read in lazy VL chunk decompression

metsw24-max:
this PR fixes a memory safety issue in the lazy variable-length chunk decompression path in blosc2.

This patch introduces:

- Checked size arithmetic to prevent overflow in trailer size calculations
- Explicit validation ensuring trailer accesses remain within `srcsize`
- Early validation during decompression context initialization to reject malformed inputs before block decoding begins

### DIFF
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -1780,8 +1780,17 @@ static int blosc_d(
     nchunk = *(int32_t*)(src + trailer_offset);
     chunk_offset = *(int64_t*)(src + trailer_offset + sizeof(int32_t));
     // Get the csize of the nblock
+    if (nblock < 0 || nblock >= context->nblocks) {
+      BLOSC_TRACE_ERROR("Invalid block index in lazy trailer.");
+      return BLOSC2_ERROR_DATA;
+    }
     int32_t *block_csizes = (int32_t *)(src + trailer_offset + sizeof(int32_t) + sizeof(int64_t));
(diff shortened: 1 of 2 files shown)


metsw24-max:
should i apply these changes?

FrancescAlted:
Again, look at them carefully, and if you think they make sense, then yes.

metsw24-max:
I looked at it carefully 
I updated the lazy VL trailer diagnostics to use more accurate wording
I also added a validation that rejects invalid VL-block chunks marked as memcpyed before they reach the decompression paths

metsw24-max:
I addressed it in the lazy block read path by adding strict validation before using trailer-derived block sizes:

- Added a bounds check for nblock to prevent invalid indexing into trailer block sizes.
- Added a sanity check for block_csize so it must be > 0 and not exceed the temporary buffer capacity used for the lazy read.
- On failure, it now returns a data error instead of proceeding to read.

This prevents oversized/invalid trailer values from driving unsafe reads into the temporary buffer.

FrancescAlted:
@copilot resolve the merge conflicts in this pull request

FrancescAlted:
I am afraid this collides with recent changes in code base.  @metsw24-max rebase please.

metsw24-max:
> I am afraid this collides with recent changes in code base. @metsw24-max rebase please.

i have rebased this branch

FrancescAlted (approved):
LGTM

PR #735, opened 2026-04-17
New shared thread pool (previous alternate)

FrancescAlted:
This is a new shared managed thread-pool that allows for much less threads
usage.

The goal of the implementation is to avoid per-context thread creation for
Blosc-managed multithreaded contexts while preserving concurrent use of
different contexts.

## Backend Model

The runtime has three internal threading backends:

- `BLOSC_BACKEND_SERIAL`
- `BLOSC_BACKEND_SHARED_POOL`
- `BLOSC_BACKEND_CALLBACK`

Selection rules:

- `nthreads <= 1` uses the serial backend.
- `nthreads > 1` and `threads_callback == NULL` uses the shared managed pool.
- `nthreads > 1` and `threads_callback != NULL` uses the callback backend.

`ctx_uses_parallel_backend()` in `blosc/context.h` is the internal predicate
for "this context is parallel" and should be used instead of inferring this
from `threads_started`.

### DIFF
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -107,6 +107,7 @@ struct blosc_job_group {
   int32_t next_output_block;
   int32_t blocks_completed;
   int32_t active_workers;
+  int32_t pending_workers;
   int32_t output_bytes;
   int giveup_code;
   int dref_not_init;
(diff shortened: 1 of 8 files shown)


FrancescAlted:
Superseeds #734

PR #736, opened 2026-04-18
Fix out-of-bounds read in frame parsing validation

metsw24-max:
This patch fixes a reader-side memory safety issue in frame parsing where chunk offsets derived from untrusted frame data were not sufficiently validated before being used in pointer arithmetic.

Malformed frames could supply crafted offset values that lead to out-of-bounds reads when accessing chunk data.

This patch introduces strict validation for offsets and chunk boundaries across all relevant parsing paths:

1. Offset validation
2. Chunk boundary validation
3. Hardened parsing paths
4. Improved error handling


FrancescAlted (approved):
LGTM

PR #737, opened 2026-04-18
Harden sframe chunk path handling

metsw24-max:
This patch hardens sparse-frame chunk file handling to prevent chunk ID aliasing and unsafe path construction.

- Added validation for sparse-frame chunk indices so values outside the valid 32-bit filename range are rejected.
- Replaced unbounded string formatting with bounded path construction and overflow-aware length checks in sparse-frame path builders.
- Reused the validated sparse-frame chunk opener in lazy-load paths to remove duplicated ad-hoc path formatting logic.
- Added a regression test that verifies out-of-range chunk IDs are rejected and do not resolve to chunk 0 via truncation.

### DIFF
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -13,6 +13,7 @@
 #include "blosc-private.h"
 #include "../plugins/codecs/zfp/blosc2-zfp.h"
 #include "frame.h"
+#include "sframe.h"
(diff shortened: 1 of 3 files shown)


metsw24-max:
i updated as per the review

FrancescAlted (approved):
LGTM

PR #743, opened 2026-04-20
Fix integer overflow in metalayer bounds

uwezkhan:
While looking at how metalayers are parsed in frame.c, I noticed that some of the bounds checks rely on adding values together (like offset + content_len). If those values are large enough, the addition can wrap around and the check may incorrectly pass.

This change updates those checks to use a safer pattern that avoids that situation. It also adds a small constant (FRAME_META_HDR_SIZE) instead of using a hardcoded value, so it’s a bit clearer what’s being checked.

I’ve applied the same fix in all similar places in frame.c and added a test that builds a malformed frame to make sure these cases are handled properly.

### DIFF
--- a/blosc/frame.c
+++ b/blosc/frame.c
@@ -1596,7 +1596,7 @@ static int get_meta_from_header(blosc2_frame_s* frame, blosc2_schunk* schunk, ui
     }
     // Go to offset and see if we have the correct marker
     uint8_t* content_marker = header + offset;
-    if (header_len < offset + 1 + 4) {
+    if (offset + FRAME_META_HDR_SIZE > header_len) {
       return BLOSC2_ERROR_READ_BUFFER;
     }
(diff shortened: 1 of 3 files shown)


FrancescAlted:
Tests are not passing. Can you have a look at this?

uwezkhan:
@FrancescAlted can you give me any suggestions on this i am trying to improve still the files are failing

FrancescAlted:
> @FrancescAlted can you give me any suggestions on this i am trying to improve still the files are failing

I think you should explain better what you are trying to fix.  My suggestion is that you start with a minimal reproducible test unit that fails; this will show clearly the issue.  Then you should start trying a minimal patch for fixing the issue, until it passes all tests locally.  Only after that, send the PR.

As this is open for too long and the use case is not that clear, I am going to close this.  Feel free to open another PR by following the guidelines above.

PR #748, opened 2026-04-21
Return -1 in case of unknown or unsupported compcode

### LINKED ISSUE #700: API contract violation in `blosc2_compcode_to_compname` on unknown compcode
On code>=6, `blosc2_compcode_to_compname` returns a non-negative code but produces a `null` compname; according to the docs, `-1` should be returned if the codec is not recognized. (not a serious issue but could be fixed easily)

I believe the issue is here: https://github.com/Blosc/c-blosc2/blob/4808890791a1cd2880d9c3f85d710fbee2c1df11/blosc/blosc2.c#L293-L294

**testcase.cpp**
```cpp
#include <cstdio>
#include <cstdlib>
extern "C" {
#include "blosc2.h"
}

int main() {
  // Pick a compcode that is not mapped/available in this build; 6 reproduces the issue here
  int code = 6;
  const char* compname = nullptr;
  int rc = blosc2_compcode_to_compname(code, &compname);
  // According to documentation, unknown/unsupported codes should return -1.
  // Observed behavior: rc == code (>=0) and compname == NULL for code 6.
  if (rc >= 0 && compname == nullptr) {
    // Demonstrate the contract violation
    fprintf(stderr, "contract violation: rc=%d, compname is NULL for code=%d\n", rc, code);
    abort();
  }
  return 0;
}
```

DimitriPapadopoulos:
From the [documentation](https://blosc.org/c-blosc2/reference/blosc1.html#c.blosc2_compcode_to_compname):
> If the compressor code is not recognized, or there is not support for it in this build, -1 is returned.

Fixes #700.

Need to add the code snippet in as a test. Not sure where to add it, perhaps a new test file?

### DIFF
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -274,20 +274,29 @@ int blosc2_compcode_to_compname(int compcode, const char** compname) {
   int code = -1;    /* -1 means non-existent compressor code */
   const char* name = NULL;
 
-  /* Map the compressor code */
-  if (compcode == BLOSC_BLOSCLZ)
+  if (compcode == BLOSC_BLOSCLZ) {
+    code = BLOSC_BLOSCLZ;
     name = BLOSC_BLOSCLZ_COMPNAME;
-  else if (compcode == BLOSC_LZ4)
+  } else if (compcode == BLOSC_LZ4) {
+    code = BLOSC_LZ4;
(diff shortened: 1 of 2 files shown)


FrancescAlted:
Thanks @DimitriPapadopoulos !

PR #749, opened 2026-04-21
guard size arithmetic to prevent overflow-driven OOB access

jmestwa-coder:
## Summary

Guard mmap read/write arithmetic to prevent integer overflow from propagating into memory operations.

## Problem

The code computed:

* `size * nitems`
* `position + n_bytes`

without validating overflow.
Overflowed values can pass bounds checks and reach memory operations with incorrect sizes, leading to out-of-bounds access.

## Proof

Example:

* `size = INT64_MAX`, `nitems = 2`
* `size * nitems` overflows and produces an invalid value
* This value can be used in subsequent checks and memory operations

## Fix

* Validate arithmetic before use:

  * check `size * nitems`
  * check `position + n_bytes`
* Abort early on invalid inputs:

  * write → returns `0`
  * read → sets `*ptr = NULL` and returns `0`

## Impact

* No change for valid inputs
* Invalid/overflow inputs are rejected before any memory access
* No API, ABI, or platform behavior changes


FrancescAlted:
This has conflicts. Please rebase.

FrancescAlted:
Still conflicts.  We have updated the mmap code quite a bit recently, so chances are that this has been addressed already. I am closing this. In case you detect a new flaw, open a new PR.

PR #764, opened 2026-05-05
Validate frame trailer length to prevent invalid offset computation

metsw24-max:
This patch adds strict validation for `trailer_len` in both file-backed and in-memory frame constructors to prevent malformed input from triggering invalid offset calculations.

#### Added an additional invariant check in:

- `frame_from_file_offset`
- `frame_from_cframe`

#### New condition:

`trailer_len <= frame_len - FRAME_HEADER_MINLEN`

#### Full validation now ensures:

- `trailer_len >= FRAME_TRAILER_MINLEN`
- `trailer_len <= INT32_MAX`
- `trailer_len <= frame_len`
- `trailer_len <= frame_len - FRAME_HEADER_MINLEN`

Invalid values are rejected early.

#### Added test_schunk_from_buffer_bad_trailer_len:

- Generates a valid contiguous frame
- Mutates only the trailer length field
- Sets it to exceed `frame_len - FRAME_HEADER_MINLEN`
- Verifies `blosc2_schunk_from_buffer` rejects the frame

### DIFF
--- a/blosc/frame.c
+++ b/blosc/frame.c
@@ -1032,7 +1032,9 @@ blosc2_frame_s* frame_from_file_offset(const char* urlpath, const blosc2_io *io,
     }
     uint32_t trailer_len;
     to_big(&trailer_len, trailer_ptr + trailer_offset, sizeof(trailer_len));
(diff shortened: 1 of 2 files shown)


FrancescAlted:
LGTM

FrancescAlted:
Thanks @metsw24-max !

PR #766, opened 2026-05-06
Fix integer overflow in metalayer bounds check

uwezkhan:
This fixes a signed integer overflow issue in metalayer parsing inside frame.c.

The previous bounds check used addition with attacker-controlled content_len values, which could overflow and bypass the validation for malformed frames with very large metadata lengths.

The fix rewrites the check using subtraction-based bounds validation, which keeps the original logic while avoiding overflow during the comparison.

The patch is intentionally minimal and only changes the affected bounds checks in:

get_meta_from_header
get_vlmeta_from_trailer

### DIFF
--- a/blosc/frame.c
+++ b/blosc/frame.c
@@ -1638,7 +1638,7 @@ static int get_meta_from_header(blosc2_frame_s* frame, blosc2_schunk* schunk, ui
     }
     // Go to offset and see if we have the correct marker
     uint8_t* content_marker = header + offset;
-    if (header_len < offset + 1 + 4) {
+    if (offset > header_len - 1 - 4) {
       return BLOSC2_ERROR_READ_BUFFER;
     }
     if (*content_marker != 0xc6) {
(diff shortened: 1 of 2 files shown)


FrancescAlted (changes requested):
The new overflow checks are good, but not quite complete. Just before them, the code still has:

```c
  if (header_len < offset + 1 + 4)
```

and similarly for trailer_len. That expression can also overflow in theory if offset is near INT32_MAX. Better would be:

```c
  if (offset > header_len - 1 - 4) {
    return BLOSC2_ERROR_READ_BUFFER;
  }

```

and same for trailer:

```c
  if (offset > trailer_len - 1 - 4) {
    return BLOSC2_ERROR_READ_BUFFER;
  }
```

A secondary concern is that tests/test_frame_malformed_metalayers.c is not a great test: if parsing unexpectedly succeeds, it prints “Unexpected” but still returns 0. So it does not actually fail the test suite. It should assert schunk2 == NULL and return failure otherwise, or be converted to the existing cutest style.

uwezkhan:
Thanks for the review.

I updated the remaining offset + 1 + 4 checks to the subtraction-based form in both header and trailer parsing paths.

I also rewrote test_frame_malformed_metalayers.c to use the existing cutest-style framework and added a strict mu_assert so the test now properly fails if malformed parsing unexpectedly succeeds.

uwezkhan:
closing this in favor of #772

PR #768, opened 2026-05-07
Fix overflow in metalayer bounds validation

uwezkhan:
This patch fixes an integer overflow issue in the metalayer parsing logic inside frame.c.

The previous bounds validation used addition with attacker-controlled values when checking metalayer offsets and content lengths. In malformed frames with very large metadata lengths, those additions could overflow and bypass the intended validation logic.

The fix updates the checks to use subtraction-based validation instead, which preserves the existing behavior for valid frames while avoiding overflow during the comparison.

The changes are intentionally small and limited to:

get_meta_from_header
get_vlmeta_from_trailer

The regression test for malformed metalayers was also updated to use the existing cutest-style assertions so the test suite properly fails if malformed parsing unexpectedly succeeds.

### DIFF
--- a/blosc/frame.c
+++ b/blosc/frame.c
@@ -1638,7 +1638,7 @@ static int get_meta_from_header(blosc2_frame_s* frame, blosc2_schunk* schunk, ui
     }
     // Go to offset and see if we have the correct marker
     uint8_t* content_marker = header + offset;
-    if (header_len < offset + 1 + 4) {
+    if (offset > header_len - 1 - 4) {
       return BLOSC2_ERROR_READ_BUFFER;
     }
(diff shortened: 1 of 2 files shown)


FrancescAlted:
Is this a duplicate of #766 ?

uwezkhan:
It started from the same investigation around bounds validation in frame parsing, but #766 is now limited to the metalayer overflow checks and regression test only.

I reverted the broader parser hardening changes there to keep the scope minimal and focused on the specific metalayer issue.

Closing this in favor of #766 to avoid duplicate review effort.

The minimal metalayer overflow fix and regression test are being tracked there now.

PR #771, opened 2026-05-09
Validate lazychunk cbytes against frame length to prevent OOB read

metsw24-max:
Fix an out-of-bounds read issue in the contiguous in-memory branch of `frame_get_lazychunk()` by validating lazychunk `cbytes` values against the frame boundaries.

Previously, the post-parse validation logic was guarded by:

```c
if (rc && ...)
```

Since `blosc2_cbuffer_sizes()` returns `0` on success, the validation path became unreachable for successfully parsed chunk headers. A crafted frame could therefore advertise an inflated `cbytes` value extending beyond the end of the frame buffer, causing downstream callers to trust an invalid chunk size and potentially read past `frame->cframe`.

This patch adds proper validation for:

* Negative chunk offsets
* Invalid offset ordering
* Integer overflow during bounds calculations
* Undersized chunk lengths
* Chunks extending beyond the frame length
* Chunk header parsing failures

### `blosc/frame.c`

* Added validation for:

  * `chunk_header_offset < 0`
  * `chunk_cbytes_offset < chunk_header_offset`
  * `chunk_cbytes_offset > frame->len`
* Added explicit error handling when `blosc2_cbuffer_sizes()` fails
* Replaced the ineffective `if (rc && ...)` validation guard with correct success-path validation
* Added overflow-safe arithmetic checks using:

  ```c
  chunk_header_offset > INT64_MAX - lazychunk_cbytes
  ```
* Added minimum valid chunk size validation using:

  ```c
  BLOSC_EXTENDED_HEADER_LENGTH
  ```

### Tests

Added `test_frame_lazychunk_malformed_cbytes.c` regression coverage for:

* Rejecting malformed frames with inflated `cbytes`
* Preventing oversized chunks from extending beyond frame boundaries
* Verifying valid lazychunk retrieval still succeeds
* Confirming successful decompression and round-trip data integrity

### DIFF
--- a/blosc/frame.c
+++ b/blosc/frame.c
@@ -2948,12 +2948,20 @@ int frame_get_lazychunk(blosc2_frame_s *frame, int64_t nchunk, uint8_t **chunk, 
 
     *chunk = frame->cframe + chunk_header_offset;
 
-    if (chunk_cbytes_offset > frame->len) {
+    if (chunk_header_offset < 0 ||
+        chunk_cbytes_offset < chunk_header_offset ||
+        chunk_cbytes_offset > frame->len) {
       BLOSC_TRACE_ERROR("Cannot read the header for chunk in the (contiguous) frame.");
       rc = BLOSC2_ERROR_READ_BUFFER;
     } else {
       rc = blosc2_cbuffer_sizes(*chunk, NULL, &lazychunk_cbytes, NULL);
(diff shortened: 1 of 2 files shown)


FrancescAlted (approved):
LGTM

FrancescAlted:
Thanks @metsw24-max !

PR #775, opened 2026-05-13
Validate negative indices in reorder_offsets

uwezkhan:
This patch fixes an out-of-bounds heap access in `blosc2_schunk_reorder_offsets`.

The function validated indices larger than `nchunks`, but did not reject negative indices from the user-provided `offsets_order` array.

As a result, negative values could be used to index heap-allocated buffers out of bounds during offset reordering.

The fix adds a lower-bound validation check to ensure indices remain within the valid range:

c id="r4vrho"
0 <= index < schunk->nchunks

A regression test was also added to verify malformed negative indices are rejected correctly.


FrancescAlted (approved):
LGTM

FrancescAlted:
Thanks @uwezkhan !

PR #783, opened 2026-06-06
size b2nd shape buffers to B2ND_MAX_DIM in ndmean and ndlz

alhudz:
Repro: store a b2nd array whose metalayer declares `ndim` in `9..16`, with the `ndmean` filter (or the `ndlz` codec), then read it back. ASAN reports a heap write past the `shape`/`chunkshape`/`blockshape` buffers in `b2nd_deserialize_meta`.
Cause: `ndmean_forward`/`ndmean_backward` and `ndlz4_compress`/`ndlz8_compress` allocate those buffers with `8` slots, but `b2nd_deserialize_meta` fills up to `B2ND_MAX_DIM` (`16`) entries straight from the metalayer. `ndmean` then indexes its `NDMEAN_MAX_DIM` stack arrays with the same unchecked `ndim`.
Fix: size the buffers to `B2ND_MAX_DIM`, as `ndcell` and the `zfp` codec already do, and reject `ndim` outside `1..NDMEAN_MAX_DIM` in `ndmean` the way `ndcell` does.

### DIFF
--- a/plugins/codecs/ndlz/ndlz4x4.c
+++ b/plugins/codecs/ndlz/ndlz4x4.c
@@ -69,9 +69,9 @@ int ndlz4_compress(const uint8_t *input, int32_t input_len, uint8_t *output, int
   }
 
   int8_t ndim;
-  int64_t *shape = malloc(8 * sizeof(int64_t));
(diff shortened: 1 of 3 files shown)


FrancescAlted:
Already addressed in #781.

PR #792, opened 2026-07-20
check chunk size before reading the flags2 byte

fr-manvi:
The flags2 byte that records whether a chunk uses variable-length blocks sits at offset 0x1e, which is inside the extended header. A Blosc1-style chunk is only BLOSC_MIN_HEADER_LENGTH bytes long and read_chunk_header() treats cbytes == 16 as valid, so such a chunk has no flags2 byte at all. The append, insert and update paths in schunk.c read chunk[BLOSC2_CHUNK_BLOSC2_FLAGS2] to test for BLOSC2_VL_BLOCKS before they call blosc2_cbuffer_sizes(), so handing a 16-byte chunk to blosc2_schunk_append_chunk() reads 15 bytes past the end of the caller's buffer. I ran into it under ASAN while pushing minimal chunks through the public append API, and then found schunk_get_chunk_flags2() has the same problem on chunks already stored in the schunk, so patching only the append site moves the same over-read to the second append instead of removing it. Routing all of them through one helper that reports flags2 as unset below BLOSC_EXTENDED_HEADER_LENGTH keeps the size lookup ahead of the read. Chunks carrying a real extended header always have the byte, so their behaviour is unchanged, and the added test covers the four entry points.

### DIFF
--- a/blosc/schunk.c
+++ b/blosc/schunk.c
@@ -889,9 +889,24 @@ static int schunk_get_chunk_nbytes(blosc2_schunk *schunk, int64_t nchunk, int32_
   return rc;
 }
 
+/* The flags2 byte lives in the extended header, at offset 0x1e.  A Blosc1-style
+   chunk is only BLOSC_MIN_HEADER_LENGTH (16) bytes long and carries no flags2,
+   so report it as unset instead of reading past the end of the buffer. */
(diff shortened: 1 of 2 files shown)


FrancescAlted:
Any opinion on the copilot suggestions?

FrancescAlted (approved):
LGTM

PR #793, opened 2026-07-23
derive special-value typesize from cbytes in getitem

fr-manvi:
A special-value chunk stores its repeated element only once, so the real element width is cbytes minus the extended header, not the typesize byte that sits in the chunk header. read_chunk_header only validates that derived width and leaves the header typesize byte unconstrained for these chunks. In _blosc_getitem the special-value branch passed context->typesize (that unchecked header byte) straight to set_values, which reads a value of that width from the chunk and replicates it, so a chunk whose typesize byte is larger than cbytes minus 32 makes set_values read past the end of the buffer and copy those bytes into the caller output. I ran into it under ASAN calling blosc2_getitem_ctx on a hand-built repeatval chunk with the value one byte wide and the typesize byte bumped to 8. blosc_d already derives the width from cbytes for the same case, so I made the getitem path derive it the same way and kept the change inside that branch. The added test builds such a chunk in an exact-size allocation and checks getitem stays within it.

### DIFF
--- a/blosc/blosc2.c
+++ b/blosc/blosc2.c
@@ -4338,14 +4338,19 @@ int _blosc_getitem(blosc2_context* context, blosc_header* header, const void* sr
     // Short-circuit for (non-lazy) memcpyed or special values
     ntbytes = nitems_bytes;
     switch (context->special_type) {
-      case BLOSC2_SPECIAL_VALUE:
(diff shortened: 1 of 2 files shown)


FrancescAlted:
Thanks @fr-manvi !

